PR #3, opened 2024-07-12
implement forecast cron job per target

Rickyxstar:
Supporting configuration for per AIScaleTarget forecast scheduling.


herrewig (approved):
We're shippin' y'all!! ⛵

PR #4, opened 2024-07-12
Use imageTag to override deployment images 

Rickyxstar:
# Why

* Currently there's no way for users to specify what specific forecast version should be run. We need to give them that control
* `image` is less accurate of a value name than `imageTag` so we'll want to level up our naming accuracy

# What's changing?

* Provide new environment variable called `MODEL_IMAGE` that enables the user to specify the desired forecast image tag
* Rename instances of `image` to the more accurately named `imageTag`


herrewig (approved):
w00t! +1 for more accurate naming

PR #12, opened 2024-07-19
HA improvements

herrewig:

The pursuit of high availability is always happening. We're always looking for ways to optimize resilience and lower the chances of issues

# What's changing?

* Give the api server 3 replicas instead of 1
* On deployment updates, ensure the metrics collector gets terminated before spinning up a new pod


charts/thoras/templates/collector/deployment.yaml
Rickyxstar, line 1: Amazing!

Rickyxstar (approved):

PR #16, opened 2024-07-26
Bumpt to 2.6.0

herrewig:

Bump chart version in prep for release


Rickyxstar (approved):
Looks great to me!

PR #18, opened 2024-08-09
Update scale.thoras.ai rule

Rickyxstar:

App changes require updating the MutatingWebhookConfiguration for `scale.thoras.ai`

# What's changing?

scale.thoras.ai MutatingWebhookConfiguration rule from `aiscaletargets/scale` to `*/scale`

### DIFF
--- a/charts/thoras/templates/api-server/webhooks.yaml
+++ b/charts/thoras/templates/api-server/webhooks.yaml
@@ -28,7 +28,7 @@ data:
 apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
 metadata:
-  name: "pods.thoras.ai"
+  name: "thoras.ai"
 webhooks:
 - name: "pods.thoras.ai"
   rules:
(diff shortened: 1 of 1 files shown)


charts/thoras/templates/api-server/webhooks.yaml
```diff
@@ -28,7 +28,7 @@ data:
 apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
 metadata:
-  name: "pods.thoras.ai"
+  name: "thoras.ai"
```
herrewig, line 31: @Rickyxstar assuming that changing the webhook name will cause us to create a new webhook instead of modifying the old, let's huddle to figure out how to ensure users who do a simple upgrade end up in a clean state with only one webhook

charts/thoras/templates/api-server/webhooks.yaml
```diff
@@ -49,10 +49,10 @@ webhooks:
     - v1
 - name: "scale.thoras.ai"
   rules:
-  - apiGroups:   ["thoras.ai"]
-    apiVersions: ["v1"]
+  - apiGroups:   ["*"]
+    apiVersions: ["*"]
     operations:  ["UPDATE"]
-    resources:   ["aiscaletargets/scale"]
+    resources:   ["*/scale"]
```
herrewig, line 55: I think we decided to be explicit with `deployment` for now, right?

charts/thoras/templates/api-server/webhooks.yaml
Rickyxstar: I reverted the name change!

charts/thoras/templates/api-server/webhooks.yaml
Rickyxstar: Updated to deployments/scale!

herrewig (approved):
LGTM!

PR #19, opened 2024-08-13
add thoras-monitor deployment

dylan-moradpour:

We are adding a new service that monitors and alerts users of various potential issues both in their infrastructure and on the thoras platform (e.g. a deployment going down).

# What's changing?

Added new deployment and corresponding values for `thoras-monitor`.

### DIFF
--- a/charts/thoras/values.yaml
+++ b/charts/thoras/values.yaml
@@ -60,4 +60,14 @@ thorasDashboard:
     memory: 100Mi
   port: 80
 
+thorasMonitor:
+  enabled: false
+  monitorCadenceSeconds: "15"
+  maxJobLifeSeconds: "600"
+  underProvisionThreshold: "0.1"
+  overProvisionThreshold: "10"
+  slackWorkspaceID: ""
+  slackChannelID: ""
+  slackWebhookID: ""
+
 thorasForecast: {}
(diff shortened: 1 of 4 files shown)


charts/thoras/values.yaml
```diff
@@ -53,4 +53,15 @@ thorasDashboard:
     memory: 100Mi
   port: 80
 
+thorasMonitor:
+  enabled: false
+  monitorCadence: "15"
```
herrewig, line 58: Can we somehow expose units either through key or value? I think users won't know what to do without at least a hint. `monitorCadenceSeconds` seems like a pretty easy way to achieve this with no app changes

charts/thoras/values.yaml
```diff
@@ -53,4 +53,15 @@ thorasDashboard:
...
+  maxJobLife: "600"
```
herrewig, line 59: Same deal as https://github.com/thoras-ai/helm-charts/pull/19/files#r1716050611 -- `maxJobLifeSeconds` would solve any confusion about what to input

charts/thoras/values.yaml
```diff
@@ -53,4 +53,15 @@ thorasDashboard:
     memory: 100Mi
   port: 80
 
+thorasMonitor:
```
herrewig, line 56: I see we're referencing `.Values.thorasMonitor.imageTag` in line 29 of the deployment but there's no default value. That seems problematic -- could we set the default value to .Values.thorasVersion?

herrewig (approved):
LGTM!

PR #20, opened 2024-08-13
Fix metrics purge entry point arguments 

Rickyxstar:

Metrics purge job pod fails on startup. 

# What's changing?

Fixed entry point arguments so that metrics purge job can start.


herrewig (approved):
LGTM!

PR #24, opened 2024-08-16
Use latest Thoras version, 2.0.0

herrewig:

We always want the latest chart to use the latest version of Thoras. Let's make sure we're defaulting to install the latest Thoras platform version, `2.0.0`

# What's changing?

Make `.Values.thorasVersion` default to `2.0.0`


Rickyxstar (approved):
Lets great!

PR #28, opened 2024-08-21
Added optional Resource Quota

Rickyxstar:

This change is being implemented to mitigate the risk of Thoras potentially consuming all available resources in a cluster in the event of a cascading cronjob failure. 

# What's changing?

An optional resource quota has been introduced to help manage resource consumption within Thoras. The following configuration options are now available:

| Key | Type | Default | Description |
| --- | --- | --- | --- |
| resourceQuota.enabled | Bool | false | Enables resource quotas within Thoras |
| resourceQuota.pods | Number | 200 | Maximum number of pods allowed|
| resourceQuota.cronjobs | Number | 200 | Maximum number of cronjobs allowed |
| resourceQuota.jobs | Number | 200 | Maximum number of jobs allowed |


herrewig (approved):
LGTM!!

PR #29, opened 2024-08-26
Wait for ES readiness before monitor start

herrewig:

In certain scenarios we may need to wait for ES to come up / initialize. Let's try waiting first before giving up!

One callout: this is the second time we do this (first is for the collector). If we end up needing it a third time, we should consider throwing this script into the container image

# What's changing?

Give ES a chance to come up before bailing


Rickyxstar (approved):
Looks good to me! 

_Why did the cargo ship get promoted? Because it really knows how to **deliver** results!_

PR #32, opened 2024-08-27
Provide ability to autocreate StorageClass for Thoras EFS volume

herrewig:

Users shouldn't have to set up an EFS `StorageClass` object in their cluster if they want to use EFS for Thoras in addition to the easiest install experience possible

# What's changing?

* Create the ability for users to automatically create a dedicated thoras `StorageClass` that's bound to the Thoras EFS volume
* Update docs
* Set a default `storageClassName` since users who want this automated probably don't want to have to think about its name

### DIFF
--- a/charts/thoras/values.yaml
+++ b/charts/thoras/values.yaml
@@ -27,7 +27,9 @@ metricsCollector:
   persistence:
     enabled: false
     volumeName: ""
-    storageClassName: ""
+    storageClassName: "efs-sc-thoras"
+    createEFSStorageClass:
+      fileSystemId: ""
   podAnnotations: {}
   collector:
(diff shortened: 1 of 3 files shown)


Rickyxstar (approved):
This is going to make a lot of people very happy! 🍝🍝🍝

charts/thoras/values.yaml
```diff
@@ -27,7 +27,9 @@ metricsCollector:
   persistence:
     enabled: false
     volumeName: ""
-    storageClassName: ""
+    storageClassName: "efs-sc-thoras"
```
herrewig, line 30: I made this a sane default since users may now be using auto-create-sc functionality of the chart

PR #33, opened 2024-08-27
Rollout app version 2.0.3

Rickyxstar:
# Purpose of Change

Thoras is rolling out version 2.0.3 of the application, which includes important updates and refinements to the suggestion logic.

# Summary of Changes

The updates will enhance the reliability and accuracy of Thoras' suggestion feature.


herrewig (approved):
Looks great! Can we try to merge https://github.com/thoras-ai/helm-charts/pull/32 first so EFS can make it into the release?

PR #37, opened 2024-09-04
Release app version 2.0.5

Rickyxstar:

Thoras deployments on spot instance based clusters have occasionally experienced elevated noise levels due to pod disruptions.

# What's changing?

To mitigate this issue, we are introducing an alert aggregation window for Thoras deployments. This window requires an error state to persist for at least two minutes before an alert is triggered. 

This change allows sufficient time for the Kubernetes scheduler to relocate Thoras pods to a healthy node in the event of a disruption. This minimizes the impact of transient errors and reduces unnecessary alerts.


herrewig (approved):
LGTM!!

PR #39, opened 2024-09-06
Use a default of 3 dashboard replicas to ensure HA

herrewig:

The Thoras dashboard should be HA which is super hard with only one replica 😂 

# What's changing?

* Allow configurable dashboard replica value
* Default to 3 for HA


Rickyxstar (approved):
Looks great to me!

PR #45, opened 2024-09-17
Provide more verbose logging for the API server

herrewig:

Customers may find more verbose logs for the API server useful. We can always adjust down the road if this is too verbose for a default

# What's changing?

* Increasing verbosity of the API server logs
* Bump version to 3.2.0


Rickyxstar (approved):
Gets great to me!! 🥇

PR #49, opened 2024-09-25
Add SLACK_WEBHOOK_URL and SLACK_ERRORS_ENABLED to helm charts

dspigarelli:

To turn on sending `error`-level logs to slack, we need to expose the environment variables.

Needed for https://github.com/thoras-ai/platform/pull/73

# What's changing?

This PR adds global and service-level configurations for `SLACK_WEBHOOK_URL` and `SLACK_ERRORS_ENABLED`. In the case of `thorasMonitor`, the webhook URL makes the passing of workspace, channel and hook IDs obsolete.

### DIFF
--- a/charts/thoras/README.md
+++ b/charts/thoras/README.md
@@ -7,7 +7,7 @@ This Helm Chart installs [Thoras](https://www.thoras.ai) onto Kubernetes.
 ![Version: 3.2.1](https://img.shields.io/badge/Version-3.2.1-informational?style=flat-square) ![AppVersion: 2.0.10](https://img.shields.io/badge/AppVersion-2.0.10-informational?style=flat-square)
 
 # Install
-Using [Helm](https://helm.sh), you can easily install and test Throas in a Kubernetes cluster by running the following:
+Using [Helm](https://helm.sh), you can easily install and test Thoras in a Kubernetes cluster by running the following:
 
 #### Add Helm repo
 First, add the repo if you haven't already done so:
@@ -57,6 +57,8 @@ helm install \
 | resourceQuota.cronjobs | Number | 200 | Maximum number of cronjobs allowed |
 | resourceQuota.jobs | Number | 200 | Maximum number of jobs allowed |
 | logLevel | String | info | Default log level |
+| slackWebhookUrl | String | "" | Slack Webhook URL destination for notifications. |
+| slackErrorsEnabled | Boolean | false | Determines if error-level logs are sent to `slackWebHookUrl` |
 
 ## Thoras Forecast
 | Key | Type | Default | Description |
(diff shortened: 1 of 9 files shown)


charts/thoras/README.md
```diff
@@ -71,6 +73,8 @@ helm install \
 | thorasOperator.limits.memory | String | 1000Mi | Thoras Operator memory limit |
 | thorasOperator.requests.cpu | String | 1000m | Thoras Operator CPU request |
 | thorasOperator.requests.memory | String | 1000Mi | Thoras Operator memory request |
+| thorasOperator.slackWebhookUrl | String | "" | Slack Webhook URL destination for notifications |
```
dspigarelli, line 76: I think it's worth allowing these to be configured at the service level, but I also welcome feedback on whether this is making it more complicated than necessary.

charts/thoras/templates/api-server/deployment.yaml
```diff
@@ -42,6 +42,10 @@ spec:
               secretKeyRef:
                 name: thoras-elastic-password
                 key: host
+          - name: SLACK_WEBHOOK_URL
+            value: {{ .Values.thorasApiServer.slackWebhookUrl | default .Values.slackWebhookUrl }}
```
dspigarelli, line 46: This is my first time editing helm templates. [The documentation](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function) seemed to favor piping things through values, but I notice that we don't do that for `default` in other places in the templates. I'm open to feedback here.

charts/thoras/templates/monitor/deployment.yaml
```diff
@@ -75,10 +75,10 @@ spec:
               secretKeyRef:
                 name: thoras-slack
                 key: webhookID
-          - name: SLACK_WORKSPACE_ID
```
dspigarelli, line 78: Do we need to worry about being backwards compatible? Seems we wouldn't want to release this until what's https://github.com/thoras-ai/platform/pull/73 is merged to `main`

charts/thoras/templates/api-server/deployment.yaml
herrewig: Nope, I imagine the way we've been doing it historically is legacy syntax. Thanks for leveling up our best practices!

charts/thoras/templates/monitor/deployment.yaml
herrewig: Yep, 100%. This would be a breaking change -- we'll just need to include instructions that explain the new value layout when we ask customers to upgrade. We'll also want to bump the major version of the chart to `3.x.x` since we're breaking (versions are cheap tho!!)

It may not always be preferable to do a hard break, but I think it makes sense for this for now

I've filed a [ticket](https://github.com/thoras-ai/platform/issues/85) to track this

dspigarelli: I bumped the version to 4.0.1

charts/thoras/templates/api-server/deployment.yaml
```diff
@@ -42,6 +42,10 @@ spec:
...
+          - name: SLACK_ERRORS_ENABLED
+            value: {{ .Values.thorasApiServer.slackErrorsEnabled | default .Values.slackErrorsEnabled | default false }}
```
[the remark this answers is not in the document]
dspigarelli: Yep, makes sense. I added a commit with your suggestion. Let me know if I misunderstood.

PR #50, opened 2024-09-27
Convert the slackWebhookUrl into a secret

dspigarelli:

When converting over to the webhook URL in #49, I didn't make it a secret

# What's changing?

I think this captures the effort in turning the `slackWebhookUrl` into a secret.


herrewig (approved):
Heck yeah!!! Let's ship itttt!!! 🚢 🚢 🚢 🚢

PR #51, opened 2024-09-27
Update AST crd and add validation

Rickyxstar:

The Thoras application will now source the values for spec.horizontal.maxReplicas, spec.horizontal.metrics, and spec.horizontal.minReplicas directly from the target's Horizontal Pod Autoscaler (HPA), eliminating the need to define these parameters within the AST configuration.

# What's changing?

Removed: The fields spec.horizontal.maxReplicas, spec.horizontal.metrics, and spec.horizontal.minReplicas have been removed from the AIScaleTarget Custom Resource Definition (CRD).
Added: A validation webhook has been introduced to ensure the integrity and correctness of the AST configuration during the creation and modification of ASTs.


charts/thoras/templates/api-server/webhooks.yaml
```diff
@@ -93,3 +93,27 @@ webhooks:
...
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["aiscaletargets"]
+      scope:       "Namespaced"
+    clientConfig:
+      service:
+        namespace: {{ .Release.Namespace }}
+        name: thoras-api-server
+        path: /validate
+      caBundle: {{ $secretCaCert | quote }}
```
dspigarelli, line 114: Oh, nifty! We could probably do that for the `SLACK_ERRORS_ENABLED` values. No need to change anything, just TIL.

charts/thoras/templates/api-server/webhooks.yaml
```diff
@@ -93,3 +93,27 @@ webhooks:
...
+    sideEffects: None
+    failurePolicy: 'Ignore'
```
dspigarelli, line 116: Just to confirm, we want to allow them to make an incorrect edit to an AST if our validator fails for some reason?

dspigarelli (approved):
LGTM! I'm not sure I understand completely all the K8s yaml being removed, but I think I get the gist of it.

charts/thoras/templates/api-server/webhooks.yaml
Rickyxstar: This is worth discussing as a group on Monday! We take this approach for all other webhooks but this might be a special case where we don't want them to continue with changes if we cannot verify them. 

On the other hand, this could stop them from applying critical fixes when modifying an AST.

herrewig (approved):
Ship it!!!! 🚢

PR #52, opened 2024-09-30
One SlackURL var to rule them all!

herrewig:

Since we have one URL at the top level, we need to reference that var!

# What's changing?

Update monitor template to reference top level Slack webhook URL var


dspigarelli (approved):
LGTM

PR #53, opened 2024-09-30
Add  v2 API service

dspigarelli:

We have a new API service layer that needs to be added to the chart for it to be released.

# What's changing?

Adding a new, `thoras-api-server-v2` deployment and service. The service is default off unless the value `thorasServerV2.enabled == true`.

### DIFF
--- a/charts/thoras/values.yaml
+++ b/charts/thoras/values.yaml
@@ -60,6 +60,16 @@ thorasApiServer:
   port: 443
   logLevel: "debug"
 
+thorasApiServerV2:
+  enabled: false
+  containerPort: 8443
+  podAnnotations: {}
+  requests:
+    cpu: 128m
+    memory: 100Mi
+  port: 443
+  logLevel: "info"
+
 thorasDashboard:
   enabled: true
   serviceAccount:
--- a/charts/thoras/templates/api-server-v2/deployment.yaml
+++ b/charts/thoras/templates/api-server-v2/deployment.yaml
@@ -1,0 +1,73 @@
+{{- if .Values.thorasApiServerV2.enabled }}
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: thoras-api-server-v2
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: {{ .Chart.Name }}
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+spec:
+  replicas: 3
+  selector:
+    matchLabels:
+      app: thoras-api-server-v2
+  template:
+    metadata:
+      labels:
+        app: thoras-api-server-v2
+        app.kubernetes.io/name: {{ .Chart.Name }}
+        helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+        app.kubernetes.io/managed-by: {{ .Release.Service }}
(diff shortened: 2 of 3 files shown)


charts/thoras/values.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 
-thorasVersion: "2.0.10"
+thorasVersion: "2.0.11"
```
dspigarelli, line 3: Do we always bump the version? Or should this only when we're ready to cut a new release?

Rickyxstar: This is when we cut a new release in the app!

Rickyxstar: This is where we vbump the helm chart:
https://github.com/thoras-ai/helm-charts/blob/d7476ed9c041b873f70c07c7f9782b79af97faf2/charts/thoras/Chart.yaml#L18

charts/thoras/values.yaml
```diff
@@ -60,6 +60,19 @@ thorasApiServer:
   port: 443
   logLevel: "debug"
 
+thorasApiServerV2:
+  enabled: false
+  containerPort: 8443
+  podAnnotations: {}
+  limits:
+    cpu: 1000m
```
herrewig, line 68: [boulder] Can we nix the cpu limit? CPU request is fine but [it's general best practice to avoid limiting CPU](https://home.robusta.dev/blog/stop-using-cpu-limits) unless there's a super specific and unique failure mode you're defending against.

charts/thoras/values.yaml
```diff
@@ -60,6 +60,19 @@ thorasApiServer:
...
+    memory: 1000Mi
+  requests:
+    cpu: 100m
+    memory: 100Mi
+  port: 443
+  logLevel: "debug"
```
herrewig, line 74: [pebble] Should we default to info instead of here so we don't have to remember to switch it back when we go to prod? The onus of setting debug can rest on the caller for now (i can help set this up in our values file), holla

Rickyxstar: Just a heads up, this limit exists for every deployment. Not just the new api service.

charts/thoras/values.yaml
Rickyxstar: Just a heads up, this limit exists for every deployment. Not just the new api service

charts/thoras/values.yaml
```diff
@@ -60,6 +60,19 @@ thorasApiServer:
   port: 443
   logLevel: "debug"
 
+thorasApiServerV2:
+  enabled: false
+  containerPort: 8443
```
herrewig, line 65: Question -- will we be mounting a TLS key pair into the pod and doing our own TLS termination in code?

Rickyxstar: Its mounted in:
https://github.com/thoras-ai/helm-charts/blob/d7476ed9c041b873f70c07c7f9782b79af97faf2/charts/thoras/templates/api-server-v2/deployment.yaml#L32

charts/thoras/values.yaml
```diff
@@ -60,6 +60,19 @@ thorasApiServer:
...
+  podAnnotations: {}
+  limits:
+    cpu: 1000m
+    memory: 1000Mi
+  requests:
+    cpu: 100m
```
herrewig, line 71: [microbes] Just sharing that I read somewhere that using powers of 2 for these types of k8s settings can actually makes things fit better in the end, on top of being more aesthetically pleasing (so `128m` in this instance) -- not sure how much BS that is but I always do it cuz I agree it looks more professional and who knows maybe it does fit better why not

herrewig (approved):
LGTM assuming we agree to nix the CPU limits

🚢 !!!

Rickyxstar (approved):
Ship it! ⚡⚡⚡

PR #54, opened 2024-10-01
Drop CPU limits; bump memory to 2Gi

dspigarelli:

We don't need to set CPU limits and increasing the memory limit will reduce future OOMs

# What's changing?

- Removing `limits.cpu` in all cases
- Bumping `limits.memory` to 2000Mi in all cases


Rickyxstar (approved):
LGTM ⚡⚡⚡

herrewig (approved):
LGTM!!!

PR #59, opened 2024-10-10
Enable API service layer

dspigarelli:

The API service layer is ready for deploying every where, not just where enabled.

# What's changing?

The API service layer is an internal service used by various components to store and retrieve data from Kubernetes and our internal storage. Initially, none of these components are using the service layer, but we'll soon be moving various components to the new service until all components are going through the layer.

### DIFF
--- a/charts/thoras/values.yaml
+++ b/charts/thoras/values.yaml
@@ -58,15 +58,14 @@ thorasApiServer:
   logLevel: "debug"
 
 thorasApiServerV2:
-  enabled: false
-  containerPort: 8443
+  containerPort: 8080
   podAnnotations: {}
   limits:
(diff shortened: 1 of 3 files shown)


charts/thoras/values.yaml
```diff
@@ -58,15 +58,14 @@ thorasApiServer:
   logLevel: "debug"
 
 thorasApiServerV2:
-  enabled: false
-  containerPort: 8443
+  containerPort: 8080
```
dspigarelli, line 61: This will change back to 8443 once the service implements TLS.

Rickyxstar (approved):
Gets great to me! ⚡

PR #60, opened 2024-10-10
Add TLS support to API service layer

dspigarelli:

To add TLS support to the API service layer, we need to pass the certificate and key file.

# What's changing?

Updating the API service layer command to include the TLS certificate and key files and updating the ports to be on 443.


herrewig (approved):
LGTM!

PR #64, opened 2024-10-16
Fix direct volume name assignment

herrewig:

If users provision a PV out-of-band from the Thoras installation, they should be able to reference that PV and be done with it.

The issue is that in that case, we'll still set the `storageClassName` field which is no longer applicable and could cause undesired behavior

# What's changing?

Make the helm chart not set `storageClassName` field on the PVC when `volumeName` is set by the user


Rickyxstar (approved):
Lets good to me!

Issue #69, opened 2024-10-18
Re-enable TLS for the API service layer


dspigarelli:
When the API service layer has TLS enabled, the dashboard fails to fetch data, and errors in the log look like this:

```
  ⨯ Error: unable to verify the first certificate
     at TLSSocket.onConnectSecure (node:_tls_wrap:1677:34)
     at TLSSocket.emit (node:events:519:28)
     at TLSSocket._finishInit (node:_tls_wrap:1076:8)
     at ssl.onhandshakedone (node:_tls_wrap:862:12)
     at TLSWrap.callbackTrampoline (node:internal/async_hooks:130:17) {
   code: 'UNABLE_TO_VERIFY_LEAF_SIGNATURE'
```

Related, the API service layer gives the following error:

```
2024/10/18 15:44:04 http: TLS handshake error from 10.128.64.179:59750: EOF
```

PR #70, opened 2024-10-18
Turn off TLS for api-v2 server

dspigarelli:

I recently switched the dashboard to use the new api service layer, but there are TLS errors that break the fetching of data for the graphs on the dashboard. This PR switches off the TLS settings until I can figure out what the issue is. I'm not sure if it's an issue with the dashboard talking to the API service layer or something else.

Example error: 
```
thoras-api-server-v2-75bb7795c4-skzw8 thoras-v2-api-server 2024/10/18 15:22:19 http: TLS handshake error from 10.128.33.155:56608: EOF
```

# What's changing?

Turning off TLS for the API service layer. I filed https://github.com/thoras-ai/helm-charts/issues/69 to turn TLS back on


Rickyxstar (approved):
LGTM

PR #84, opened 2024-10-24
Add forecast_buffer_percentage field

herrewig:

Users should have a way to define a "safety buffer" that gets added to their usage forecasts so that they can have increased confidence that there won't be disruption.

# What's changing?

Add `spec.mode.forecast_buffer_percentage` field to the `AIScaleTarget` CRD. Can be an integer or string


Rickyxstar (approved):
Looks great!

dspigarelli (approved):

PR #85, opened 2024-10-29
Add agent daemon set

dspigarelli:

The Thoras agent needs to run on every node in the cluster, giving it a chance to query eBPF and other metrics so we create the service map portion of the reasoning map.

# What's changing?

Adding a daemon set that runs the `thoras-agent` with elevated privileges.


charts/thoras/templates/agent/daemonset.yaml
```diff
@@ -0,0 +1,56 @@
...
+        app.kubernetes.io/name: {{ .Chart.Name }}
+        helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+        app.kubernetes.io/managed-by: {{ .Release.Service }}
+        app.kubernetes.io/instance: {{ .Release.Name }}
+      {{- with .Values.thorasAgent.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+    spec:
+      serviceAccountName: thoras-collector
```
dspigarelli, line 29: It seems we only have a couple service accounts. The `thoras-collector` is used by several other deployments and I figured I'd reuse it as well. We can add an additional, agent-specific service if we feel that reuse isn't appropriate.

charts/thoras/templates/agent/daemonset.yaml
```diff
@@ -0,0 +1,56 @@
...
+            value: {{ default .Values.logLevel .Values.thorasAgent.logLevel }}
+        name: thoras-agent
+        resources:
+          limits:
+            memory: 128Mi
+          requests:
+            cpu: 100m
+            memory: 128Mi
+      # drawn from https://github.com/kubeservice-stack/kubservice-charts/blob/6be7b9fb081abd2255dc83854dbed2d7d11e0973/charts/kubeservice-ebpf-exporter/values.yaml#L67-L69
+      tolerations:
```
dspigarelli, line 46: I'm not 100% sure what tolerations are, but it seems like something we need. I drew this setting from another eBPF-related daemon set, link above.

charts/thoras/values.yaml
```diff
@@ -95,3 +95,6 @@ thorasMonitor:
 
 thorasForecast:
   requestCpu:
+
+thorasAgent:
+  enabled: false
```
dspigarelli, line 100: Not enabled by default. This will likely be enabled once we have the basic, beta-version completed.

charts/thoras/templates/agent/daemonset.yaml
Rickyxstar: Do you know what api calls the agent will need to make against the k8s api?

dspigarelli: I don't think it will be making any calls, for now. But this was needed to get access to `thoras-secret-registry` so the images could be pulled.

Rickyxstar (approved):
LGTM ⚡⚡⚡

PR #92, opened 2024-11-01
Run reasoning api via gunicorn in prod

herrewig:

Since python's single-threaded, Flask by itself is not sufficient for non-development use-cases. We need a WSGI server to act as a reverse proxy so concurrent HTTP calls are managed properly. Luckily this is a well-established pattern and easy to do

# What's changing?

Run reasoning API via gunicorn

### DIFF
--- a/charts/thoras/templates/reasoning-api/deployment.yaml
+++ b/charts/thoras/templates/reasoning-api/deployment.yaml
@@ -33,7 +33,15 @@ spec:
       - image: {{ .Values.imageCredentials.registry }}/thoras-forecast:{{ default .Values.thorasVersion .Values.thorasReasoningApi.imageTag }}
         imagePullPolicy: "{{ .Values.imagePullPolicy }}"
         name: thoras-reasoning-api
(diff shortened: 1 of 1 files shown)


charts/thoras/templates/reasoning-api/deployment.yaml
```diff
@@ -33,7 +33,15 @@ spec:
       - image: {{ .Values.imageCredentials.registry }}/thoras-forecast:{{ default .Values.thorasVersion .Values.thorasReasoningApi.imageTag }}
         imagePullPolicy: "{{ .Values.imagePullPolicy }}"
         name: thoras-reasoning-api
-        command: ["/usr/local/bin/reasoning_api"]
+        command:
+          - "gunicorn"
+          - "--workers"
+          - "4"
```
herrewig, line 39: Hard-coded for now so we can ship the smallest useful thing but we'll want to do something more clever at some point

dylan-moradpour (approved):

PR #93, opened 2024-11-01
Add environment variable for reasoning API endpoint

dspigarelli:

In order to talk to the internal reasoning API, we need to plumb an environment variable through to the service.

# What's changing?

Adding a `SERVICE_REASONING_API_URL` that points to the reasoning service. It's prefixed with `SERVICE_` because that's how the [huma cli](https://huma.rocks/features/cli/#passing-options) expects environment variables to be prefixed. 

Also fixing a couple other environment variables that were previously being handled by [envconfig](https://github.com/kelseyhightower/envconfig) but weren't changed when we switched them to huma.


herrewig (approved):
LGTM!

PR #94, opened 2024-11-04
Added reasoning metrics list

Rickyxstar:

Preparing the AIScaleTarget CRD for a new reasoning feature. 

# What's changing?

Added new `spec.reasoning.metrics` list that allows the user to configure where their custom metrics live and how to query them. 

Example:
```yaml
apiVersion: thoras.ai/v1
kind: AIScaleTarget
metadata:
  name: game-server
  namespace: battle-all-nite
spec:
  horizontal:
    mode: autonomous
  model:
    forecast_blocks: 60m
    forecast_cron: '*/15 * * * *'
    mode: balanced
  reasoning:
    metrics:
    - name: http_requests_total
      query: http_requests_total{environment=~"staging|testing|development",method!="GET"}
      connector: prometheus
  scaleTargetRef:
    kind: Deployment
    name: game-server
  vertical:
    mode: recommendation
```

### DIFF
--- a/charts/thoras/templates/crd/aiscaletarget.yaml
+++ b/charts/thoras/templates/crd/aiscaletarget.yaml
@@ -274,7 +274,28 @@ spec:
                       x-kubernetes-int-or-string: true
                     forecast_buffer_percentage:
                       x-kubernetes-int-or-string: true
-
+                reasoning:
+                  type: object
+                  required:
+                    - metrics
+                  properties:
+                    metrics:
+                      type: array
+                      items:
+                        type: object
+                        required:
+                            - type
+                            - name
+                            - query
+                        properties:
+                          type:
+                            type: string
+                            pattern: ^(prom)$
+                          name:
+                            type: string
+                            pattern: ^[a-zA-Z0-9_-]+$
+                          query:
+                            type: string
             status:
               type: object
               properties:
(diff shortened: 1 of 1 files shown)


charts/thoras/templates/crd/aiscaletarget.yaml
```diff
@@ -105,8 +105,28 @@ spec:
...
+                  type: object
+                  required:
+                    - metrics
+                  properties:
+                    metrics:
+                      type: array
+                      items:
+                        type: object
+                        required:
+                            - type
```
herrewig, line 118: bikeshedding time: what about `"provider"` instead of `type`? Seems more accurate

charts/thoras/templates/crd/aiscaletarget.yaml
```diff
@@ -105,8 +105,28 @@ spec:
...
+                            - name
+                            - query
+                        properties:
+                          type:
+                            type: string
+                            pattern: ^(prom)$
```
herrewig, line 124: These regexes not being in quotes scares me a bit given how YAML has a specific semantics around certain symbols -- would it hurt to throw quotes around these pattern fields?

charts/thoras/templates/crd/aiscaletarget.yaml
```diff
@@ -274,7 +274,28 @@ spec:
...
+                      items:
+                        type: object
+                        required:
+                            - connector
+                            - name
+                            - query
+                        properties:
+                          connector:
+                            type: string
+                            pattern: '^(prometheus)$'
```
herrewig, line 293: I love we can enforce this all the way at the crd level

herrewig (approved):
CRD whisperer!! 🚢

PR #97, opened 2024-11-05
Move securityContext from pod to container

dspigarelli:

In order to collect eBPF metrics, the agent container must be running with `securityContext: privileged: true`, not the pod.

# What's changing?

I'm moving the `securityContext` block to the container from the pod. Without doing this, we get permission errors when starting the agent, like this:

```
time=2024-11-05T17:31:22.310Z level=ERROR msg="Failed to initialize collector" err="loading eBPF objects: field TcpCleanupRbuf: program tcp_cleanup_rbuf: load program: operation not permitted (MEMLOCK may be too low, consider rlimit.RemoveMemlock)"
panic: loading eBPF objects: field TcpCleanupRbuf: program tcp_cleanup_rbuf: load program: operation not permitted (MEMLOCK may be too low, consider rlimit.RemoveMemlock)
```

### DIFF
--- a/charts/thoras/templates/agent/daemonset.yaml
+++ b/charts/thoras/templates/agent/daemonset.yaml
@@ -41,13 +41,13 @@ spec:
           requests:
             cpu: 100m
             memory: 128Mi
+        securityContext:
+          # the agent requires elevated privileges to collect ebpf traces
+          privileged: true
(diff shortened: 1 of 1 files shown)


herrewig (approved):
LGTM

PR #98, opened 2024-11-05
External metrics crd

Rickyxstar:

This enhancement introduces a new feature that allows Thoras users to specify metrics for the Thoras Forecaster to use for predictions and recommendations, independent of the metrics used by the target HPA for scaling.

# What's changing?

A new `spec.scale_metrics` property has been added to the AIScaleTarget CRD. Example configuration:

```
---
apiVersion: thoras.ai/v1
kind: AIScaleTarget
metadata:
  name: nginx-target
  namespace: keda
spec:
  horizontal:
    mode: recommendation
  model:
    forecast_buffer_percentage: 10%
    forecast_cron: '*/10 * * * *'
    mode: balanced
  scale_metrics:
  - external:
      metric:
        name: s0-aws-cloudwatch
        selector:
          matchLabels:
            scaledobject.keda.sh/name: rtb-bidder
      target:
        averageValue: "1"
        type: AverageValue
    type: External
  scaleTargetRef:
    kind: Deployment
    name: avg-deployment
```

### DIFF
--- a/charts/thoras/templates/crd/aiscaletarget.yaml
+++ b/charts/thoras/templates/crd/aiscaletarget.yaml
@@ -89,6 +89,167 @@ spec:
                                 x-kubernetes-int-or-string: true
                         required:
                           - name
+
+
+
+                scale_metrics:
+                  type: array
+                  items:
+                    type: object
+                    properties:
+                      type:
+                        type: string
+                        pattern: '^(Object|Resource|Pods|External)$'
+                      object:
+                        type: object
+                        required:
+                          - metric
+                          - describedObject
+                          - target
+                        properties:
+                          metric:
+                            type: object
+                            required:
+                              - name
+                            properties:
+                              name:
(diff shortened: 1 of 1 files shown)


charts/thoras/templates/crd/aiscaletarget.yaml
```diff
@@ -89,6 +89,173 @@ spec:
...
+                                  - type: integer
+                                  - type: string
+                                x-kubernetes-int-or-string: true
+                              averageValue:
+                                anyOf:
+                                  - type: integer
+                                  - type: string
+                                x-kubernetes-int-or-string: true
+                              value:
+                                anyOf:
```
herrewig, line 147: [dust] According to chatgpt we only need `x-kubernetes-int-or-string` and not both that + anyOf

herrewig (approved):
LGTM!

PR #100, opened 2024-11-06
Enable secret-by-reference

herrewig:

A common pattern that some customers expect is the ability to reference existing secrets instead of passing in sensitive values to the helm chart for secret population.

# What's changing?

* Make it possible to reference an existing registry secret
* Make it possible to reference an existing Slack URL secret
* Add unit tests

# Notes

* I re-ordered some env vars to make the Slack env variables in a consistent order. Shouldn't hurt anything
* Implementing Prometheus endpoint env var for reasoning-api directly after this

### DIFF
--- a/charts/thoras/templates/api-server-v2/deployment.yaml
+++ b/charts/thoras/templates/api-server-v2/deployment.yaml
@@ -38,8 +38,6 @@ spec:
         name: thoras-v2-api-server
         command: ["/app/api-server", "-p", {{ .Values.thorasApiServerV2.containerPort | quote }}]
         env:
-          - name: SERVICE_REASONING_API_URL
-            value: "http://thoras-reasoning-api.thoras.svc.cluster.local"
           - name: "ELASTICSEARCH_URL"
             valueFrom:
               secretKeyRef:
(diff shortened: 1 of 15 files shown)


charts/thoras/templates/api-server-v2/deployment.yaml
```diff
@@ -38,8 +38,6 @@ spec:
         name: thoras-v2-api-server
         command: ["/app/api-server", "-p", {{ .Values.thorasApiServerV2.containerPort | quote }}]
         env:
-          - name: SERVICE_REASONING_API_URL
```
herrewig, line 41: I moved this so the env field we're testing is in the same array index across all deployments . This let's us to have one helm unit test instead of many and the env var ordering shouldn't matter

charts/thoras/templates/dashboard/deployment.yaml
```diff
@@ -52,11 +62,6 @@ spec:
           {{- end }}
           - name: "MODEL_IMAGE"
             value: {{ .Values.imageCredentials.registry }}/thoras-forecast:{{ default .Values.thorasVersion .Values.thorasForecast.imageTag }}
-          - name: SLACK_WEBHOOK_URL
```
herrewig, line 55: I moved this so the env field we're testing is in the same array index across all deployments . This let's us to have one helm unit test instead of many and the env var ordering shouldn't matter

charts/thoras/templates/dashboard/rbac.yaml
```diff
@@ -1,5 +1,22 @@
 {{- if and .Values.thorasDashboard.rbac.create .Values.thorasDashboard.enabled }}
 ---
+apiVersion: v1
```
herrewig, line 3: I moved this so the env field we're testing is in the same array index across all rbac.yaml files . This let's us to have one helm unit test instead of many and the env var ordering shouldn't matter

Rickyxstar (approved):
LGTM ⚡⚡⚡

PR #103, opened 2024-11-06
Add configurable prometheus base URL for reasoning api

herrewig:

* Customers need the ability to configure their prometheus base URL for us to ingest prome metrics

# What's changing?

* Give customers the ability to configure prometheus base endpoint via  `thorasReasoningApi.connectors.prometheus.baseUrl` value

# Callouts

* This is a smallest useful thing -- we talked about probably needing something more sophisticated for managing connector config in the long term


Rickyxstar (approved):
LGTM ⚡⚡⚡!

PR #108, opened 2024-11-12
Migrate to go collector

Rickyxstar:

Migrating to a newer better metrics collector 

# What's changing?

Removing old metrics collector image in favor of a newer one.

### DIFF
--- a/charts/thoras/templates/collector/deployment.yaml
+++ b/charts/thoras/templates/collector/deployment.yaml
@@ -70,17 +70,16 @@ spec:
           runAsUser: 1000
           runAsGroup: 0
         {{- end }}
-      - image: {{ .Values.imageCredentials.registry }}/thoras-collector:{{ default .Values.thorasVersion .Values.metricsCollector.collector.imageTag }}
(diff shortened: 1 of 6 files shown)


dspigarelli (approved):
:shipit:

charts/thoras/templates/collector/deployment.yaml
```diff
@@ -70,17 +70,16 @@ spec:
...
         name: {{ .Values.metricsCollector.collector.name }}
         command: ["/bin/sh", "-c"]
         args:
           - |
-            # Don't start app until ES is available
             retries=0
             max_retries=48
             until [ $retries -eq $max_retries ]; do
-              if curl -s --connect-timeout 5 --fail -XGET ${ES_HOST}; then
+              if wget --spider -T 5 -q ${ELASTICSEARCH_URL}; then
```
herrewig, line 82: I'm totally good w/the change just curious why we swapped this out

dspigarelli: I think it' because the `alpine:3.20` image doesn't have `curl` by default but it does have `wget`. We could have installed curl like we are for the previous [metrics collector images](https://github.com/thoras-ai/platform/blob/c7991c653daa8d1d3c7546f1aaf3a790760caa77/packages/thoras-collector/Dockerfile#L30)

PR #109, opened 2024-11-19
Make PVC requests.storage configurable

herrewig:

Users need to be able to specify `spec.resources.requests.storage` on their PVC -- especially in case they're using a fixed storage size backend (such as AWS EBS)

# What's changing?

* Make `spec.resources.requests.storage` configurable
* Default to "3Gi"
* Add unit tests
* Update documentation


Rickyxstar (approved):
LGTM ⚡⚡⚡

PR #111, opened 2024-11-20
Make helm unit tests more flexible with environment variable checks

dspigarelli:

When adding a new environment variable for some services, I was burned by tests that were too rigid in the expected order of environment variables.

# What's changing?

Updating some tests to care less about the index of environment variables and more about what the environment variables look like, using [jsonpath](https://github.com/json-path/JsonPath)

### DIFF
--- a/charts/thoras/tests/secrets_collector_test.yaml
+++ b/charts/thoras/tests/secrets_collector_test.yaml
@@ -8,10 +8,10 @@ tests:
       slackWebhookUrlSecretRefKey: "url"
     asserts:
       - equal:
-          path: spec.template.spec.containers[1].env[2].valueFrom.secretKeyRef.name
+          path: $..spec.containers[1].env[?(@.name == 'SLACK_WEBHOOK_URL')].valueFrom.secretKeyRef.name
           value: "frou-frou"
       - equal:
-          path: spec.template.spec.containers[1].env[2].valueFrom.secretKeyRef.key
+          path: $..spec.containers[1].env[?(@.name == 'SLACK_WEBHOOK_URL')].valueFrom.secretKeyRef.key
           value: "url"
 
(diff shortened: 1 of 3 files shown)


charts/thoras/tests/secrets_collector_test.yaml
```diff
@@ -8,20 +8,20 @@ tests:
       slackWebhookUrlSecretRefKey: "url"
     asserts:
       - equal:
-          path: spec.template.spec.containers[1].env[2].valueFrom.secretKeyRef.name
+          path: $..spec.containers[1].env[?(@.name == 'SLACK_WEBHOOK_URL')].valueFrom.secretKeyRef.name
           value: "frou-frou"
       - equal:
-          path: spec.template.spec.containers[1].env[2].valueFrom.secretKeyRef.key
+          path: $..spec.containers[1].env[?(@.name == 'SLACK_WEBHOOK_URL')].valueFrom.secretKeyRef.key
           value: "url"
```
dylan-moradpour, line 15: how would these paths both point to the same value?

dylan-moradpour: the paths look the same to me - maybe Im interpreting the syntax wrong?

dylan-moradpour (approved):
LGTM! Left a non blocking question about how these tests work - I'm sure its just my lack of understanding not a problem in the change.

charts/thoras/tests/secrets_collector_test.yaml
dspigarelli: I might not be understanding the question, but the first is comparing the `.name` value, and the second is looking at the `.key`. Am I misunderstanding your question?

dylan-moradpour: My window was small and I didnt see the `.key` and `.name` at the end...my bad!

PR #112, opened 2024-11-21
Default to single replica for services

herrewig:

The best practice for open source helm charts is to default deployments to 1 replica because:

- Letting the user decide what they need is more aligned with a user-driven config approach, which is more ideal
- It's less expensive
- Avoids premature scaling
- Is easier to understand and navigate for new users

# What's changing?

Default replicas to 1 for all `Deployments`


dspigarelli (approved):
:shipit:

PR #113, opened 2024-11-21
Added rbac for v2 api

Rickyxstar:

The v2 api server will need to be able to access kubernetes resources.

# What's changing?

A service account with a corresponding cluster role has been created and assigned to the v2 API server to grant the necessary resource access.

### DIFF
--- a/charts/thoras/templates/api-server-v2/rbac.yaml
+++ b/charts/thoras/templates/api-server-v2/rbac.yaml
@@ -1,0 +1,112 @@
+{{- if and .Values.thorasApiServerV2.rbac.create  }}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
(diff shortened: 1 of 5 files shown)


charts/thoras/templates/api-server-v2/rbac.yaml
```diff
@@ -0,0 +1,112 @@
...
+  resources:
+  - jobs
+  verbs:
+  - '*'
+- apiGroups:
+  - apps
+  resources:
+  - deployments
+  verbs:
+  - patch
```
dspigarelli, line 63: For @herrewig, this is a mostly a copy of the `dashboard` role. This `patch` is what's doing the magic.

dspigarelli (approved):

herrewig (approved):
🚢 !!

PR #118, opened 2024-11-25
Update ast metric

Rickyxstar:

This change addresses user requests for the ability to exclude metrics from AIScale while retaining the functionality of Horizontal Pod Autoscaling (HPA) based on those same metrics.  This allows for more granular control over resource management and scaling.

# What's changing?

The `exclude_metrics` field has been added to the `spec.horizontal` section of the configuration, allowing users to specify metrics to exclude from AIScale.  The supported metric types are: `External`, `Resource`, and `Object`.  An example configuration is shown below:

```yaml
    exclude_metrics:
      - type: External
        name: aws-alb-latency
      - type: Resource
        name: cpu
      - type: Object
        name: requests-per-second
```

Furthermore, the `spec.scale_metrics` field has been relocated to `spec.horizontal` for improved configuration organization and clarity.

### DIFF
--- a/charts/thoras/templates/crd/aiscaletarget.yaml
+++ b/charts/thoras/templates/crd/aiscaletarget.yaml
@@ -214,48 +214,60 @@ spec:
                                     type: string
                                   averageUtilization:
                                     type: string
-                  vertical:
-                    type: object
-                    required:
-                      - mode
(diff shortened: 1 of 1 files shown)


herrewig (approved):
LGTM!

PR #120, opened 2024-12-02
Added timescaledb

Rickyxstar:

Thoras will be migrating off of Elasticsearch.

# What's changing?

Added TimescaleDB. 

TimescaleDB is disabled by default.

### DIFF
--- a/charts/thoras/templates/collector/deployment.yaml
+++ b/charts/thoras/templates/collector/deployment.yaml
@@ -36,10 +36,12 @@ spec:
         command: ["/bin/sh", "-c"]
         args:
           - |
-            chown -R 1000:0 /usr/share/elasticsearch/data
-            rm -f /usr/share/elasticsearch/data/node.lock
+            mkdir -p /var/lib/share/elasticsearch/data
+            find /var/lib/share/elasticsearch -not -name postgresql -not -name data -mindepth 1 -maxdepth 1 -exec mv {} /var/lib/share/elasticsearch/data \;
+            chown -R 1000:0 /var/lib/share/elasticsearch/data
+            rm -f /var/lib/share/elasticsearch/data/node.lock
         volumeMounts:
-          - mountPath: /usr/share/elasticsearch/data
+          - mountPath: /var/lib/share/elasticsearch
             name: elastic-search-data
       {{- end }}
       containers:
@@ -57,6 +59,8 @@ spec:
             value: "true"
           - name: bootstrap.memory_lock
             value: "false"
+          - name: path.data
+            value: /var/lib/share/elasticsearch/data
           - name: "ELASTIC_PASSWORD"
             valueFrom:
               secretKeyRef:
@@ -64,12 +68,35 @@ spec:
                 key: password
         {{- if .Values.metricsCollector.persistence.enabled }}
         volumeMounts:
-          - mountPath: /usr/share/elasticsearch/data
+          - mountPath: /var/lib/share/elasticsearch
             name: elastic-search-data
         securityContext:
           runAsUser: 1000
           runAsGroup: 0
         {{- end }}
+      {{- if .Values.metricsCollector.timescale.enabled }}
+      - image: timescale/timescaledb:latest-pg17
+        name: timescaledb
+        imagePullPolicy: "{{ .Values.imagePullPolicy }}"
+        ports:
+        - containerPort: 5432
+        env:
+          - name: "POSTGRES_PASSWORD"
+            valueFrom:
+              secretKeyRef:
+                name: thoras-timescale-password
+                key: password
+          - name: "PGDATA"
+            value: /var/lib/share/postgresql
+        {{- if .Values.metricsCollector.persistence.enabled }}
+        volumeMounts:
+          - mountPath: /var/lib/share
+            name: elastic-search-data
+        securityContext:
+          runAsUser: 1000
+          runAsGroup: 0
+        {{- end }}
+      {{- end }}
       - image: {{ .Values.imageCredentials.registry }}/services:{{ default .Values.thorasVersion .Values.metricsCollector.collector.imageTag }}
         imagePullPolicy: "{{ .Values.imagePullPolicy }}"
         name: {{ .Values.metricsCollector.collector.name }}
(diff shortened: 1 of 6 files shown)


charts/thoras/templates/collector/deployment.yaml
```diff
@@ -36,10 +36,12 @@ spec:
         command: ["/bin/sh", "-c"]
         args:
           - |
-            chown -R 1000:0 /usr/share/elasticsearch/data
-            rm -f /usr/share/elasticsearch/data/node.lock
+            mkdir -p /var/lib/share/elasticsearch/data
+            find /var/lib/share/elasticsearch -not -name postgresql -not -name data -mindepth 1 -maxdepth 1 -exec mv {} /var/lib/share/elasticsearch/data \;
```
herrewig, line 40: I'm sure I'm just not fully understanding , but if the `find` command here is meant to move the files out of the legacy filepath (`/usr/share/elasticsearch/`), shouldn't that be the first argument to `find`? It looks like it's finding in the directory it just created in the line above and not the old directory path

dspigarelli: The `find` command is relative to the mount point, which also changed.

Aside: the mount point moved from `/usr/share/elasticsearch` because that's actually where elastic is mounting it's other files, including `bin`.

charts/thoras/templates/collector/deployment.yaml
```diff
@@ -57,19 +59,44 @@ spec:
...
         volumeMounts:
-          - mountPath: /usr/share/elasticsearch/data
+          - mountPath: /var/lib/share/elasticsearch
             name: elastic-search-data
         securityContext:
           runAsUser: 1000
           runAsGroup: 0
         {{- end }}
+      {{- if .Values.metricsCollector.timescale.enabled }}
+      - image: timescale/timescaledb:latest-pg17
```
herrewig, line 78: This is great for now but may I ask we file a ticket to copy this this image to our internal registry in line with our other upstream images (like elasticsearch). It might just be easier to copy it over now 

This is useful to:

* Prevent us from getting rate limited by dockerhub which is much less lenient than Artifact Registry
* Provides traffic / download telemetry
* Protects Thoras in the event that timescale dockerhub account gets popped and a malicious actor publishes an infected version by the same name

dspigarelli: Oh, yeah, @Rickyxstar mentioned this but then we forgot to do it.

charts/thoras/templates/collector/secret.yaml
```diff
@@ -10,3 +10,14 @@ data:
...
   name: thoras-elastic-password
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: thoras-timescale-password
+data:
+  {{- $secretObj := (lookup "v1" "Secret" .Release.Namespace "thoras-timescale-password") }}
+  {{- $secretData := (get $secretObj "data") | default dict }}
+  {{- $esPassword := (get $secretData "password") | default ((.Values.metricsCollector.timescale.password | default $newEsPassword) | b64enc) }}
```
herrewig, line 21: [boulder] It looks like we're reusing the `newEsPassword` that's declared at the top of the file for the ES secret yaml document, so in the use-case where someone installs Thoras from scratch and doesn't specify new passwords (one of the most likely scenarios),  Elasticsearch and Postgres will get set assigned the same randomly generated password. That seems like something we should avoid if it's easy to do

Can we just declare a new dedicated random password variable that's independent from `newEsPassword` in the above document? Like `{{- $newPgPassword := randAlphaNum 16 }}` or something in between line 13 and 14 -- just repeating the pattern . Shouldn't require much thought

I"d also highly recommend changing `esPassword` we use here to `pgPassword` or something to prevent confusion or weird variable collisions from the ES yaml document above this one in the same file

charts/thoras/templates/collector/service.yaml
```diff
@@ -11,3 +11,16 @@ spec:
     targetPort: {{ .Values.metricsCollector.search.containerPort }}
   selector:
     app: metrics-collector
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ .Values.metricsCollector.timescale.name }}
```
herrewig, line 18: [pebble] Any reason we're giving the user the option to name this? [I don't believe we've historically done that for other components](https://github.com/thoras-ai/helm-charts/blob/main/charts/thoras/templates/dashboard/service.yaml#L5) -- feels a bit like YAGNI and unnecessary config complexity to me personally. I'm not super strong on it tho -- it's working software

charts/thoras/tests/__snapshot__/collector_deployment_test.yaml.snap
```diff
@@ -9,6 +9,8 @@ Default containers should match snapshots:
         value: "true"
       - name: bootstrap.memory_lock
         value: "false"
+      - name: path.data
```
herrewig, line 12: It's funny to me their env var convention doesn't include any indication what it's associated with. ES age showin hard through here (not that PG is younger 😂 but at least their env vars comply with best practices)

charts/thoras/tests/collector_deployment_test.yaml
```diff
@@ -40,5 +40,14 @@ tests:
       - equal:
           path: spec.template.spec.containers[0].volumeMounts
           value:
-            - mountPath: /usr/share/elasticsearch/data
+            - mountPath: /var/lib/share/elasticsearch
               name: elastic-search-data
+  - it: Should create timescale container when enabled
```
herrewig, line 52: [pebble] Pebble-level feedback so it's your choice, but for these I think it's useful to have another test that asserts it's not there when the value is omitted or false. I might be more concerned about accidentally running PG when we didn't mean to versus not running PG when we meant to

herrewig (approved):
🚢

charts/thoras/tests/collector_deployment_test.yaml
herrewig: ignore me! I see we're basically doing this above

charts/thoras/templates/collector/secret.yaml
dspigarelli: Done!

PR #123, opened 2024-12-05
Add timescale migration init container using goose

dspigarelli:

We need an init container that will migrate timescale database versions.

# What's changing?

I'm adding an init container that will wait for timescaledb and then migrate the database versions. I put the init container on the v2 service because (a) the migrations will likely be associated with a API service layer change and (b) because the init container can't be in the same deployment as the timescaledb container because it requires timescaledb to be up and running before it can run.

### DIFF
--- a/charts/thoras/templates/api-server-v2/deployment.yaml
+++ b/charts/thoras/templates/api-server-v2/deployment.yaml
@@ -32,6 +32,27 @@ spec:
         - name: tls
           secret:
             secretName: thoras-api-server-cert
+      {{- if .Values.metricsCollector.timescale.enabled }}
+      initContainers:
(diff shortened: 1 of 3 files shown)


charts/thoras/templates/api-server-v2/deployment.yaml
```diff
@@ -32,6 +32,27 @@ spec:
         - name: tls
           secret:
             secretName: thoras-api-server-cert
+      {{- if .Values.metricsCollector.timescale.enabled }}
```
dspigarelli, line 35: It feels kind icky referring to the metrics collector settings from the api service layer deployment. Open to feedback here.

PR #124, opened 2024-12-05
Use dbmate to migrate database

dspigarelli:

We need an init container that will migrate timescale database versions.

# What's changing?

I'm adding an init container that will wait for timescaledb and then migrate the database versions. I put the init container on the v2 service because (a) the migrations will likely be associated with a API service layer change and (b) because the init container can't be in the same deployment as the timescaledb container because it requires timescaledb to be up and running before it can run.

### DIFF
--- a/charts/thoras/templates/api-server-v2/deployment.yaml
+++ b/charts/thoras/templates/api-server-v2/deployment.yaml
@@ -32,12 +32,36 @@ spec:
         - name: tls
           secret:
             secretName: thoras-api-server-cert
+      {{- if .Values.metricsCollector.timescale.enabled }}
+      initContainers:
+      - name: migrate-database
+        image: {{ .Values.imageCredentials.registry }}/services:{{ default .Values.thorasVersion .Values.thorasApiServerV2.migration.imageTag }}
+        command: ["/bin/sh", "-c"]
+        args:
+          - |
+            until /app/dbmate --migrations-dir /app/migrations/ migrate
+            do
+              echo "Waiting for timescaledb...";
+              sleep 10
+            done
+        env:
+          - name: DATABASE_URL
+            valueFrom:
+              secretKeyRef:
+                name: thoras-timescale-password
+                key: host
+      {{- end}}
       containers:
       - image: {{ .Values.imageCredentials.registry }}/services:{{ default .Values.thorasVersion .Values.thorasApiServerV2.imageTag }}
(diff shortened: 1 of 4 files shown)


charts/thoras/templates/collector/secret.yaml
```diff
@@ -21,4 +21,4 @@ data:
   {{- $secretData := (get $secretObj "data") | default dict }}
   {{- $pgPassword := (get $secretData "password") | default ((.Values.metricsCollector.timescale.password | default $newPgPassword) | b64enc) }}
   password: {{ $pgPassword | quote }}
-  host: {{ (printf "postgres://postgres:%s@%s:%d" ($pgPassword | b64dec) .Values.metricsCollector.timescale.name (.Values.metricsCollector.timescale.containerPort | int)) | b64enc | quote }}
+  host: {{ (printf "postgres://postgres:%s@%s:%d?sslmode=disable" ($pgPassword | b64dec) .Values.metricsCollector.timescale.name (.Values.metricsCollector.timescale.containerPort | int)) | b64enc | quote }}
```
dspigarelli, line 24: Clears the following error:

```
Error: pq: SSL is not enabled on the server
```

charts/thoras/templates/api-server-v2/deployment.yaml
```diff
@@ -32,12 +32,36 @@ spec:
...
+                name: thoras-timescale-password
+                key: host
+      {{- end}}
       containers:
       - image: {{ .Values.imageCredentials.registry }}/services:{{ default .Values.thorasVersion .Values.thorasApiServerV2.imageTag }}
         imagePullPolicy: "{{ .Values.imagePullPolicy }}"
         name: thoras-v2-api-server
         command: ["/app/api-server", "-p", {{ .Values.thorasApiServerV2.containerPort | quote }}]
         env:
+          - name: DATABASE_URL
```
dspigarelli, line 60: We aren't actually using the timescale database yet, so this is once we do start using it.

charts/thoras/templates/collector/secret.yaml
dspigarelli: Also, see https://github.com/amacneil/dbmate?tab=readme-ov-file#postgresql

Rickyxstar (approved):
LGTM ⚡⚡⚡

PR #127, opened 2024-12-16
Add prometheus api URL to service layer environment

dspigarelli:

For the service layer to query prometheus, it needs the URL to prometheus

# What's changing?

Adding the `SERVICE_PROMETHEUS_API_URL` environment variable which points to the prometheus base url

### DIFF
--- a/charts/thoras/templates/api-server-v2/deployment.yaml
+++ b/charts/thoras/templates/api-server-v2/deployment.yaml
@@ -59,6 +59,8 @@ spec:
             value: "{{ .Values.thorasApiServerV2.slackErrorsEnabled | default .Values.slackErrorsEnabled }}"
           - name: "SERVICE_LOGLEVEL"
             value: {{ default .Values.logLevel .Values.thorasApiServerV2.logLevel }}
+          - name: "SERVICE_PROMETHEUS_API_URL"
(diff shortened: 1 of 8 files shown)


charts/thoras/templates/api-server-v2/deployment.yaml
```diff
@@ -59,6 +59,8 @@ spec:
             value: "{{ .Values.thorasApiServerV2.slackErrorsEnabled | default .Values.slackErrorsEnabled }}"
           - name: "SERVICE_LOGLEVEL"
             value: {{ default .Values.logLevel .Values.thorasApiServerV2.logLevel }}
+          - name: "SERVICE_PROMETHEUS_API_URL"
+            value: {{ .Values.thorasReasoningApi.connectors.prometheus.baseUrl }}
```
dspigarelli, line 63: Is it okay to refer to the `thorasReasoningApi` values from other deployments? Or should I be specifying this value separately for the API server layer, even though it should point to the same thing?

dspigarelli: Or maybe the prometheus endpoint URL should be a secret that both can refer to?

herrewig (approved):
Let's goooo!!!

PR #129, opened 2024-12-18
Operator image change

Rickyxstar:

Migrating operator to new services image.

# What's changing?

operator image  and command updated to use new services image.


herrewig (approved):
🚢 🚢 🚢 🚢 🚢

PR #130, opened 2024-12-20
Send postgres url to api-v2

dspigarelli:

The service layer needs the postgres/timescale DB DSN passed to it a certain way so it can connect to postgres to store metrics.

# What's changing?

Changing the environment variable to align with how [huma](https://huma.rocks/features/cli/#passing-options) expects it to be passed.


Rickyxstar (approved):
LGTM ⚡⚡⚡

PR #136, opened 2024-12-27
added init to start of metrics collector

Rickyxstar:

Elasticsearch should be initialized with indices before starting the metrics collector 

# What's changing?

Added init script to start of metrics collector


herrewig (approved):
LGTM!

dylan-moradpour (approved):

PR #137, opened 2024-12-27
Make default enabled timescale DB

parka1226:
TimescaleDB is always configured to run even when no timescale-related values are present in the config. 

# What's changing?
1. Remove timescale container validation
2. Update the unit testings accordingly
3. Update default timescale input passing in the values.yml

### DIFF
--- a/charts/thoras/templates/collector/deployment.yaml
+++ b/charts/thoras/templates/collector/deployment.yaml
@@ -76,7 +76,6 @@ spec:
           runAsUser: 1000
           runAsGroup: 0
         {{- end }}
-      {{- if .Values.metricsCollector.timescale.enabled }}
       - image: {{ .Values.imageCredentials.registry }}/timescaledb:{{ .Values.metricsCollector.timescale.imageTag }}
         name: timescaledb
         imagePullPolicy: "{{ .Values.imagePullPolicy }}"
@@ -98,7 +97,6 @@ spec:
           runAsUser: 1000
           runAsGroup: 0
         {{- end }}
-      {{- end }}
       - image: {{ .Values.imageCredentials.registry }}/services:{{ default .Values.thorasVersion .Values.metricsCollector.collector.imageTag }}
         imagePullPolicy: "{{ .Values.imagePullPolicy }}"
         name: {{ .Values.metricsCollector.collector.name }}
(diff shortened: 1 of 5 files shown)


charts/thoras/templates/collector/deployment.yaml
```diff
@@ -98,7 +97,6 @@ spec:
           runAsUser: 1000
           runAsGroup: 0
         {{- end }}
-      {{- end }}
```
dspigarelli, line 101: [boulder] There's another check against `.Values.metricsCollector.timescale.enabled` on line 122 that needs to be removed.

charts/thoras/tests/secrets_collector_test.yaml
```diff
@@ -8,20 +8,20 @@ tests:
...
+          path: $..spec.containers[2].env[?(@.name == 'SLACK_WEBHOOK_URL')].valueFrom.secretKeyRef.key
           value: "url"
 
   - it: Points all apps to default slack secret, if no existing secret provided provided
     templates:
       - collector/deployment.yaml
     asserts:
       - equal:
-          path: $..spec.containers[1].env[?(@.name == 'SLACK_WEBHOOK_URL')].valueFrom.secretKeyRef.name
+          path: $..spec.containers[2].env[?(@.name == 'SLACK_WEBHOOK_URL')].valueFrom.secretKeyRef.name
```
dspigarelli, line 22: [sand] Instead of simply updating indices, I'd suggest making these tests more resilient by constructing a [jsonpath query](https://github.com/json-path/JsonPath) that fetches the expected container, like this:

```suggestion
          path: $..spec.containers[?(@.name == 'thoras-collector')].env[?(@.name == 'SLACK_WEBHOOK_URL')].valueFrom.secretKeyRef.name
```

charts/thoras/templates/collector/deployment.yaml
parka1226: Thanks for checking this out! I'll push these changes.

charts/thoras/tests/secrets_collector_test.yaml
dspigarelli: Ditto for other instances in this file that use hard-coded indices.

charts/thoras/templates/collector/deployment.yaml
herrewig: +1 Mr eagle eye over here 👁️ lol, I would have totally missed that!

herrewig (approved):
Approved contingent on addressing @dspigarelli 's  [comment](https://github.com/thoras-ai/helm-charts/pull/137/files#r1898714200)

Amazing job working through 10000000000 new layers of abstraction to figure this stuff out, Abby!!

dspigarelli (approved):

PR #138, opened 2024-12-30
Add separate service account for agent

dspigarelli:

The thoras agent should use a different service account than the metrics collector.

# What's changing?

When I created the agent, instead of creating a proper RBAC role and service account for it, I reused the same service account from the metrics collector. This was tech debt that this PR now fixes.


Rickyxstar (approved):
LGTM ⚡⚡⚡

PR #139, opened 2024-12-31
Pass SERVICE_CLUSTER _NAME environment variable from value.yaml

parka1226:
Update our helm chart to allow customers to pass in the name of the cluster in a way that propagates the cluster name to an environment variable in the services application

# What's changing?
- Customers can define the name of their cluster in the helm chart installation config (values.yaml)
- Adding unit test that asserts this behavior works in the helm chart

### DIFF
--- a/charts/thoras/values.yaml
+++ b/charts/thoras/values.yaml
@@ -1,5 +1,7 @@
 ---
 thorasVersion: "4.1.0"
+cluster: 
+  name: ""
 
 imageCredentials:
   registry: "us-east4-docker.pkg.dev/thoras-registry/platform"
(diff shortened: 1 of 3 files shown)


charts/thoras/values.yaml
```diff
@@ -96,6 +96,8 @@ thorasApiServerV2:
     memory: 100Mi
   port: 80
   logLevel: "info"
+  cluster:
```
herrewig, line 99: [[pebble]](https://www.netlify.com/blog/2020/03/05/feedback-ladders-how-we-encode-code-reviews-at-netlify/) this 100% works and I definitely wouldn't say my suggestion is blocking, but I wouldn't be too surprised if other components may need access to the cluster name one day

If you wanted to set up the structure to imply that the variable is component-agnostic (ie not just limited to `api server v2`), you could pull it up into its own top-level variable instead of nesting it underneath `thorasApiServerV2` -- like `.Values.cluster.name`)

This could/would encourage other components to consume the variable , not just the api server

I don't feel too strongly about this and it's something we can change down the road if we decide to.

100% your call!

herrewig (approved):
Let's goooooooo!!! 🔥

PR #141, opened 2025-01-02
Fix app version

Rickyxstar:

Type in values file 

# What's changing?

Bumped app version


Rickyxstar:
Merging for pragmatism! @herrewig

herrewig:
Omg thank you!! Sorry about that fam!!

PR #142, opened 2025-01-03
494 Create end-to-end scaffolding go app for monitor-v2

parka1226:
create the the ability to deploy new app monitor-v2 via helm chart.
(https://github.com/thoras-ai/platform/issues/494)

# What's changing?

There's a new top-level config block in our helm chart called thorasMonitorV2.enabled
When this variable is set to true, a new Kubernetes Deployment is created in the Thoras installation
There are unit tests that assert no deployment is created when the variable is false and vice versa

### DIFF
--- a/charts/thoras/templates/monitor-v2/deployment.yaml
+++ b/charts/thoras/templates/monitor-v2/deployment.yaml
@@ -1,0 +1,32 @@
+{{- if .Values.thorasMonitorV2.enabled }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: thoras-monitor-v2
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: {{ .Chart.Name }}
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: thoras-monitor-v2
+  template:
+    metadata:
+      labels:
+        app: thoras-monitor-v2
+        app.kubernetes.io/name: {{ .Chart.Name }}
+        helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+        app.kubernetes.io/managed-by: {{ .Release.Service }}
+        app.kubernetes.io/instance: {{ .Release.Name }}
+    spec:
+      serviceAccountName: thoras-monitor-v2
+      containers:
+        - name: thoras-monitor-v2
+          image: {{ .Values.imageCredentials.registry }}/services:{{ default .Values.thorasVersion .Values.thorasMonitorV2.imageTag }}
+          imagePullPolicy: "{{ .Values.imagePullPolicy }}"
+          command: ["/app/monitor"]
+{{- end }}
--- a/charts/thoras/templates/monitor-v2/rbac.yaml
+++ b/charts/thoras/templates/monitor-v2/rbac.yaml
@@ -1,0 +1,92 @@
+{{- if .Values.thorasMonitorV2.enabled }}
+---
+apiVersion: v1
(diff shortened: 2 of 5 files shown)


charts/thoras/templates/monitor-v2/deployment.yaml
```diff
@@ -0,0 +1,31 @@
...
+  template:
+    metadata:
+      labels:
+        app: thoras-monitor-v2
+        app.kubernetes.io/name: {{ .Chart.Name }}
+        helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+        app.kubernetes.io/managed-by: {{ .Release.Service }}
+        app.kubernetes.io/instance: {{ .Release.Name }}
+    spec:
+      serviceAccountName: thoras-monitor-v2
```
herrewig, line 26: This won't work without us defining a new `ServiceAccount` object

While defining a new one is probably the technically correct right approach, I recommend we stick with using the `thoras-operator` account for now since a) we're [already doing that](https://github.com/thoras-ai/helm-charts/blob/main/charts/thoras/templates/monitor/deployment.yaml#L32) and b) the monitor pod probably needs the same permissions as the operator anyway. I think it's fair to shoot for parity of how the `monitoring` component works today since they need the same perms

Alternatively, you could duplicate the [operator](https://github.com/thoras-ai/helm-charts/blob/main/charts/thoras/templates/operator/rbac.yaml) ServiceAccount but I'm not sure it's worth it right now especially given we're not planning on turning on monitorV2 Yet™

Your call! But either way I think we'll need to change it

parka1226: Sounds good! I can stick with the same account to keep the same process.

charts/thoras/templates/monitor-v2/deployment.yaml
```diff
@@ -0,0 +1,31 @@
...
+      containers:
+        - name: thoras-monitor-v2
+          image: {{ .Values.imageCredentials.registry }}/thoras-monitor-v2:{{ default .Values.thorasVersion .Values.thorasMonitorV2.imageTag }}
```
dspigarelli, line 29: [pebble] All of our current go services use the same `services` image. Unless there's a good reason to build them separately, I'd recommend using the same pattern here:
```suggestion
          image: {{ .Values.imageCredentials.registry }}/services:{{ default .Values.thorasVersion .Values.thorasMonitorV2.imageTag }}
```

Assuming you go this route, you'll also need to specify a `command: [...]` entry point. Otherwise the default `/app/api-server` will be started, which is probably not your intent.

parka1226:
> serviceAccountName: thoras-monitor-v2

charts/thoras/templates/monitor-v2/rbac.yaml
```diff
@@ -0,0 +1,57 @@
...
+  namespace: {{ .Release.Namespace }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: thoras-monitor-v2
+subjects:
+- kind: ServiceAccount
+  name: thoras-monitor-v2
+  namespace: {{ .Release.Namespace }}
+---
```
herrewig, line 38: Woah it looks like you discovered a pre-existing bug!!

I was about to comment that the ClusterRoleBinding is duplicated , and that's true, but it looks like it's duplicated in the original thoras-monitor rbac.yaml file as well!

I think we're going to want to:

1. replace the duplicate second document with one more thing: a copy/paste/modify of the operator ClusterRole, here: https://github.com/thoras-ai/helm-charts/blob/main/charts/thoras/templates/component-clusterrole.yaml
2. Delete the duplicate document in the original thoras-monitor rbac.yaml
3. We'll also probably want to reduce the permissions now that I'm seeing the original operator ClusterRole. We basically want to go for a ReadOnly permission set. I'm super happy to help with that!

We can work with you on Monday to help figure out what that should be!

Thanks so much @parka1226 !!

herrewig (approved):
🚢 🚢 🚢 🚢

herrewig (approved):
Incredible job @parka1226 !! Ship it!!

PR #143, opened 2025-01-03
Give agent service account permission to query K8s for deployments and pods

dspigarelli:

In order to map IP addresses to deployments, we need to be able to query the k8s API to list the deployments and pods for each deployment.

# What's changing?

Updating the agent rbac rules to allow for `get/list` against any API

### DIFF
--- a/charts/thoras/templates/agent/daemonset.yaml
+++ b/charts/thoras/templates/agent/daemonset.yaml
@@ -60,4 +60,5 @@ spec:
       # preempts running Pods
       # priorityClassName: important
       terminationGracePeriodSeconds: 30
+      hostNetwork: true
 {{- end }}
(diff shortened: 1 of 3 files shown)


charts/thoras/templates/agent/daemonset.yaml
```diff
@@ -60,4 +60,5 @@ spec:
       # preempts running Pods
       # priorityClassName: important
       terminationGracePeriodSeconds: 30
+      hostNetwork: true
```
dspigarelli, line 63: The agent doesn't need it's on IP address and can use that of the host, similar to how other prometheus agents work.

charts/thoras/values.yaml
```diff
@@ -132,7 +132,7 @@ thorasAgent:
   serviceAccount:
     name: thoras-agent
   frequency: 15
-  containerPort: 9100
+  containerPort: 9101
```
dspigarelli, line 135: If we're going to use host networking, we don't want to conflict with other prometheus endpoints on the host.

Rickyxstar (approved):
LGTM ⚡⚡⚡

PR #144, opened 2025-01-07
Restore missing ClusterRoleBinding for the thoras-collector

dspigarelli:

The `thoras-operator -> thoras-operator` ClusterRoleBinding was removed in #142 erroneously and needs to be restored.

# What's changing?

This PR restores the ClusterRoleBinding removed in #142 and moves the `thoras-collector` ClusterRoleBinding to the collector's RBAC files.


herrewig (approved):
Thanks for everything @dspigarelli 🚢

PR #145, opened 2025-01-07
Fix broken unit test for monitor-v2 deployment

dspigarelli:

When running the tests locally (not in CI, for some reason), the following error is generated:

```
 FAIL  Monitor-V2       charts/thoras/tests/monitor_v2_deployment_test.yaml
        - Should create deployment when thorasMonitorV2.enabled is false

                - asserts[0] `notExists` fail
                        Template:       thoras/templates/monitor-v2/deployment.yaml
```

This PR fixes that error by checking for documents instead of using `notExists`

# What's changing?

Also, this PR fixes the incorrect YAML warning:
Screenshot 2025-01-07 at 2 42 05 PM

### DIFF
--- a/charts/thoras/tests/monitor_v2_deployment_test.yaml
+++ b/charts/thoras/tests/monitor_v2_deployment_test.yaml
@@ -13,12 +13,11 @@ tests:
       - equal:
           path: spec.template.spec.containers[0].name
           value: thoras-monitor-v2
-  - it: Should create deployment when thorasMonitorV2.enabled is false
+  - it: Should not create deployment when thorasMonitorV2.enabled is false
     template: monitor-v2/deployment.yaml
     set:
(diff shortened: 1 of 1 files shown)


herrewig:
FYI @parka1226 no action and it's no big deal at all but for your awareness: it looks like the version of helm unittest plugin @dspigarelli and I use on our local machines is different than the latest version that gets installed in our builds (and maybe on your machine), resulting in inconsistent behavior. The build appears to use `0.7.1` but david and I have `0.6.3` installed which cause the test to fail when I checked out `main`. I think it has to do w/the fact the unit test api is changing for `notExists` to replace `isNull`

Either way, this change should make the test work in a way that will work on everyones' machines!

Again, no action, just FYI!

herrewig (approved):
LGTM

PR #148, opened 2025-01-10
Updating env variable to SERVICE

parka1226:
Fixing bug with taking the correct prefix environment variables via huma cli. 
related to https://github.com/thoras-ai/helm-charts/pull/147

# What's changing?
Updating OPERATOR to SERVICE prefix


herrewig (approved):
🔥

PR #152, opened 2025-01-14
Automate timescale password setting

herrewig:

We need to automate setting the `postgres` user's password 

# What's changing

We detect this is the first time we're initializing timescaledb, set the password


herrewig:
Fixing the unit test now

dspigarelli (approved):
LGTM

PR #156, opened 2025-01-17
Update collector environment variables

dspigarelli:

The collector's environment variables are changing because of our underlying CLI setup.

# What's changing?

Updates to the environment variables and parameters to the collector `prune` and `collect` commands.

### DIFF
--- a/charts/thoras/templates/collector/cronjob.yaml
+++ b/charts/thoras/templates/collector/cronjob.yaml
@@ -44,8 +44,14 @@ spec:
                     key: host
               - name: DATABASE_URL
                 value: "$(DATABASE_HOST)/thoras?sslmode=disable"
+              - name: SERVICE_POSTGRESQL_DSN
+                value: "$(DATABASE_HOST)/thoras?sslmode=disable"
+              - name: "SERVICE_LOGLEVEL"
+                value: {{ default .Values.logLevel .Values.metricsCollector.purge.logLevel }}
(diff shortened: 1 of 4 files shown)


charts/thoras/templates/collector/cronjob.yaml
```diff
@@ -30,22 +30,19 @@ spec:
...
-                    name: thoras-timescale-password
-                    key: password
               - name: DATABASE_HOST
                 valueFrom:
                   secretKeyRef:
                     name: thoras-timescale-password
                     key: host
-              - name: DATABASE_URL
+              - name: SERVICE_POSTGRESQL_DSN
                 value: "$(DATABASE_HOST)/thoras?sslmode=disable"
```
parka1226, line 39: Just to make sure I'm understanding this correctly. I see we are using `timescaleRepo := repositories.NewTimescaleRepository(os.Getenv("DATABASE_URL"))` on the platform/services/cmd/metrics-collector/main.go. Are we planning to change on the platform side too or is this change has been updated?

parka1226 (approved):
LGTM!

charts/thoras/templates/collector/cronjob.yaml
dspigarelli: Discussed offline. The environment var was changed in https://github.com/thoras-ai/platform/pull/587

PR #157, opened 2025-01-17
Wait for postgres before starting api server

dspigarelli:

The API service layer should wait for timescale/postgres instead of crashing during start.

# What's changing?

Adding an init container that waits for timescale/postgres to be available before handing off to the main container.

Should cut down on errors in our slack channels, like this:

```
error
failed to connect to user=postgres database=thoras:
    172.20.118.24:5432 (timescale): dial error: dial tcp 172.20.118.24:5432: connect: connection refused
    172.20.118.24:5432 (timescale): dial error: dial tcp 172.20.118.24:5432: connect: connection refused
```

### DIFF
--- a/charts/thoras/templates/api-server-v2/deployment.yaml
+++ b/charts/thoras/templates/api-server-v2/deployment.yaml
@@ -32,6 +32,33 @@ spec:
         - name: tls
           secret:
             secretName: thoras-api-server-cert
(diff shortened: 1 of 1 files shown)


herrewig (approved):
TIL `pg_isready`

Ship it!!!

PR #159, opened 2025-01-21
Add Ray cluster to chart

dspigarelli:

We want to use [Ray](https://github.com/ray-project/ray) to process various ML/AI workloads.

# What's changing?

- Adding a chart dependency on https://ray-project.github.io/kuberay-helm/
- Adding a basic ray autoscaling setup

# Future Work

- switch from the stock ray worker image to our own image that has our particular workloads


dspigarelli:
Closing for now.

PR #162, opened 2025-01-22
Update slack error reporting-related env vars

dspigarelli:

The slack error reporting-related environment variables need to be updated for our services.

# What's changing?

We changed several environment variables around sending error messages to slack, including those for the collector, operator, monitor-v2 and the agent.

### DIFF
--- a/charts/thoras/templates/monitor-v2/deployment.yaml
+++ b/charts/thoras/templates/monitor-v2/deployment.yaml
@@ -25,8 +25,25 @@ spec:
     spec:
       serviceAccountName: thoras-monitor-v2
       containers:
-        - name: thoras-monitor-v2
-          image: {{ .Values.imageCredentials.registry }}/services:{{ default .Values.thorasVersion .Values.thorasMonitorV2.imageTag }}
-          imagePullPolicy: "{{ .Values.imagePullPolicy }}"
-          command: ["/app/monitor"]
+      - name: thoras-monitor-v2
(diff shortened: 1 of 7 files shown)


charts/thoras/templates/monitor-v2/deployment.yaml
```diff
@@ -25,8 +25,25 @@ spec:
...
+              secretKeyRef:
+            {{- if and .Values.slackWebhookUrlSecretRefName .Values.slackWebhookUrlSecretRefKey }}
+                name: {{ .Values.slackWebhookUrlSecretRefName }}
+                key: {{ .Values.slackWebhookUrlSecretRefKey }}
+            {{- else }}
+                name: thoras-slack
+                key: webhookUrl
+            {{- end }}
+          - name: SERVICE_SLACK_ERRORS_ENABLED
+            value: "{{ .Values.thorasMonitor.slackErrorsEnabled | default .Values.slackErrorsEnabled }}"
```
parka1226, line 46: This file is under monitor-v2 and is this for Values.thorasMonitorV2 or .Values.thorasMonitor?

parka1226 (approved):
I have one question to confirm and everything else LGTM!

charts/thoras/templates/monitor-v2/deployment.yaml
dspigarelli: Good question. I was using `thorasMonitor` because ideally customers would get the same logging behavior from the v2 monitor as they did from the v1 monitor without needing to specify another value. But I could make it another fallback.

parka1226: That makes sense. Thank you for the clarification!

PR #163, opened 2025-01-27
Remove duplicate SERVICE_CLUSTER_NAME env var

dspigarelli:

In moving around some environment variables, I introduced a duplicate `SERVICE_CLUSTER_NAME` for the collector deployment, which would trigger a helm chart apply error.

# What's changing?

- Dropping the duplicate `SERVICE_CLUSTER_NAME` env var for the collector
- Also re-adding the `DATABASE_URL` environment variable removed in #156 because the collector still depends on it for migrations.

### DIFF
--- a/charts/thoras/tests/__snapshot__/collector_deployment_test.yaml.snap
+++ b/charts/thoras/tests/__snapshot__/collector_deployment_test.yaml.snap
@@ -83,12 +83,12 @@ Default containers should match snapshots:
           secretKeyRef:
             key: host
             name: thoras-timescale-password
+      - name: DATABASE_URL
+        value: $(DATABASE_HOST)/thoras?sslmode=disable
       - name: SERVICE_POSTGRESQL_DSN
         value: $(DATABASE_HOST)/thoras?sslmode=disable
       - name: SERVICE_LOGLEVEL
         value: info
-      - name: SERVICE_CLUSTER_NAME
-        value: ""
     image: us-east4-docker.pkg.dev/thoras-registry/platform/services:dev
     imagePullPolicy: IfNotPresent
     name: thoras-collector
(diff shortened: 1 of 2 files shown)


dylan-moradpour (approved):

charts/thoras/tests/__snapshot__/collector_deployment_test.yaml.snap
```diff
@@ -83,12 +83,12 @@ Default containers should match snapshots:
           secretKeyRef:
             key: host
             name: thoras-timescale-password
+      - name: DATABASE_URL
+        value: $(DATABASE_HOST)/thoras?sslmode=disable
```
herrewig, line 87: @dspigarelli I could be wrong butI think this is supposed to be : `value: ${DATABASE_HOST}/thoras?sslmode=disable`

$() in bash is a subshell command execution, not env var interpolation. I think the value below it would fail too, can we fix that? (assuming I'm not wrong about this)

```
          - name: SERVICE_POSTGRESQL_DSN
            value: "$(DATABASE_HOST)/thoras?sslmode=disable"
```

dspigarelli: I'm using the conventions from [here](https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/) which use `$()` and not `${}`. I could be mistaken, but this isn't executed by bash.

herrewig: Wow , all my k8s experience doesn't seem to amount for much  lol, ignore me!

herrewig (approved):
LGTM!

PR #164, opened 2025-01-28
Use generated AST CRD

Rickyxstar:

To prioritize maintainability and alignment with Kubernetes tools as well as improve the user experience of working with ASTs  we need to migrate to a generated AST CRD.

# What's changing?

The legacy hand built AST CRD has been replaced with a controller-tools generated CRD that will improve usability as well as input validation.

### DIFF
--- a/charts/thoras/templates/crd/aiscaletarget.yaml
+++ b/charts/thoras/templates/crd/aiscaletarget.yaml
@@ -2,335 +2,668 @@
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.17.1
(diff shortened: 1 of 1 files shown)


herrewig (approved):
Yo dawg heard you like yaml 
image

charts/thoras/templates/crd/aiscaletarget.yaml
```diff
@@ -2,335 +2,668 @@
...
+  group: thoras.ai
   names:
     kind: AIScaleTarget
     listKind: AIScaleTargetList
     plural: aiscaletargets
     shortNames:
-      - ast
-  group: thoras.ai
+    - ast
+    singular: aiscaletarget
```
herrewig, line 16: TIL!

PR #166, opened 2025-02-10
Pass artifact storage path to api service layer

dspigarelli:

The service layer needs to know where to store artifacts and other blobs.

# What's changing?

- Associate the service layer with the PVC
- Pass the storage path to the service layer via environment variable

### DIFF
--- a/charts/thoras/templates/api-server-v2/deployment.yaml
+++ b/charts/thoras/templates/api-server-v2/deployment.yaml
@@ -32,6 +32,11 @@ spec:
         - name: tls
           secret:
             secretName: thoras-api-server-cert
+        {{- if .Values.metricsCollector.persistence.enabled }}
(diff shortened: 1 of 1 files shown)


charts/thoras/templates/api-server-v2/deployment.yaml
```diff
@@ -32,6 +32,11 @@ spec:
         - name: tls
           secret:
             secretName: thoras-api-server-cert
+        {{- if .Values.metricsCollector.persistence.enabled }}
```
dspigarelli, line 35: I'm using the `metricsCollector` values because I'm not 100% sure the effect of moving it will have on customers. I suspect that we'll want customers to pass this in differently in the future.

parka1226 (approved):
LGTM!

PR #167, opened 2025-02-12
Add POSTGRES_PASSWORD to collector environment

dspigarelli:

Restoring the `POSTGRES_PASSWORD` removed erroneously in #156 so that the collector can set the correct password on startup.


herrewig (approved):
🍐 🚢

PR #168, opened 2025-02-13
Run chart releaser for release** branches

dspigarelli:

In order to tag hotfix releases, we need to run the chart releaser when the version changes on any branch prefixed with `release` 

# What's changing?

Updating our github release action to include branches prefixed with `release`


herrewig (approved):
Let's gooooo!

PR #165, opened 2025-01-30
Huma expects SERVICE_LOGLEVEL not SERVICE_LOG_LEVEL

herrewig:
# Why are we making this change?

Huma expects SERVICE_LOGLEVEL not SERVICE_LOG_LEVEL

# What's changing?

Switch loglevel env var to Huma-expected SERVICE_LOGLEVEL

PR #169, opened 2025-02-14
Remove duplicate SERVICE_LOG_LEVEL

dspigarelli:

Fixes the following error: 

```49046 warnings.go:70] spec.template.spec.containers[2].env[6]: hides previous definition of "SERVICE_LOG_LEVEL", which may be dropped when using apply```

Introduced in https://github.com/thoras-ai/helm-charts/pull/165

# What's changing?

Remove the duplicate `SERVICE_LOG_LEVEL` on the collector


Rickyxstar (approved):
LGTM!

PR #170, opened 2025-02-14
Move admission web hooks to operator

Rickyxstar:

Moving legacy admission web hooks to service image and running them in the operator. 

# What's changing?

Move admission web hooks from v1 api server to operator. 
Removed v1 api server.


dspigarelli (approved):
:shipit:

PR #173, opened 2025-02-19
Don't build release branches

dspigarelli:

We shouldn't immediately build branches prefixed with `release` but should be able to manually trigger the workflow via the github UI.


herrewig (approved):
🚢

PR #174, opened 2025-02-21
Create and chown timescale directory

dspigarelli:

On a fresh install, timescale/postgres goes into a crashbackoff loop because it doesn't have perms to make the data dir

# What's changing?

Mounting and running `mkdir -p` and `chown` in an init container when starting the metrics collector.


Rickyxstar (approved):
LGTM! ⚡

PR #178, opened 2025-02-24
Add exclude_metrics back into AST CRD

dspigarelli:

The `exclude_metrics` was dropped from the AST CRD in https://github.com/thoras-ai/helm-charts/pull/164 when it shouldn't have been.


dylan-moradpour (approved):

Rickyxstar (approved):
Nice catch!

PR #181, opened 2025-02-28
Added Nginx reverse proxy to dashboard exposing api server

Rickyxstar:

To streamline communication between the frontend and the API server, we are exposing the API server on the same port as the dashboard. This eliminates the need for a Backend-for-Frontend (BFF) layer, allowing the frontend to make direct HTTP requests to the API server.

# What's changing?

An Nginx container has been added to the dashboard pod, acting as a reverse proxy for both the dashboard and the API server on the same port. This ensures seamless integration while maintaining a simplified architecture.

### DIFF
--- a/charts/thoras/tests/__snapshot__/dashboard_service_test.yaml.snap
+++ b/charts/thoras/tests/__snapshot__/dashboard_service_test.yaml.snap
@@ -9,7 +9,7 @@ Default matches snapshot:
       ports:
         - port: 80
           protocol: TCP
-          targetPort: 3000
+          targetPort: 80
       selector:
         app: thoras-dashboard
       type: ClusterIP
--- a/charts/thoras/values.yaml
+++ b/charts/thoras/values.yaml
@@ -108,6 +108,9 @@ thorasDashboard:
     create: true
   podAnnotations: {}
   containerPort: 3000
+  nginxContainerPort: 80
+  nginx:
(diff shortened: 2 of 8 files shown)


charts/thoras/values.yaml
```diff
@@ -108,6 +108,9 @@ thorasDashboard:
     create: true
   podAnnotations: {}
   containerPort: 3000
+  nginxContainerPort: 80
+  nginx:
+    imageTag: "alpine"
```
herrewig, line 113: Did we push this to our registry?

charts/thoras/tests/__snapshot__/dashboard_service_test.yaml.snap
```diff
@@ -9,7 +9,7 @@ Default matches snapshot:
       ports:
         - port: 80
           protocol: TCP
-          targetPort: 3000
+          targetPort: 80
```
herrewig, line 12: Isn't this supposed to be 4000? It looks like below `nginxContainerPort: 4000` which is what gets assigned to `targetPort: {{ .Values.thorasDashboard.nginxContainerPort }}` in dashboard service. I could be wrong!

herrewig (approved):
LGTM!

charts/thoras/values.yaml
Rickyxstar: Pushed!

PR #182, opened 2025-03-03
Allow queries-per-second overrides

dspigarelli:

We need a way to override the default k8s rate limiting "queries per second" or QPS. Future work to introduce short-window caching of target k8s queries.


Rickyxstar (approved):
LGTM! ⚡

PR #184, opened 2025-03-04
Add SERVICE_TIMESCALE_PRIMARY environment variable

dspigarelli:

We need a feature flag for enabling timescale as the primary data source instead of Elastic.


herrewig (approved):
LGTM!

PR #186, opened 2025-03-05
Use camel-casing for queriesPerSecond

dspigarelli:

We prefer `camelCase` to `snake_casing` for value overrides.


herrewig (approved):
LGTM!

PR #187, opened 2025-03-05
Add prod config to dashboard

Rickyxstar:

We are implementing reverse proxying for our API server within the dashboard pod. To support this, the dashboard configuration needs to be updated accordingly.

# What's changing?

- Added /config.json to the Nginx configuration.
- Verified functionality using Minikube.


herrewig (approved):
LGTM but for my own edification why do we return that empty api_base_url json?

Rickyxstar:
> LGTM but for my own edification why do we return that empty api_base_url json?

When you set the base URL to an empty string (""), the browser interprets it as a relative URL. This means that any request made using this base URL will inherit the protocol (http or https), hostname, and port from the current page's URL. Essentially, the browser keeps using the same origin (protocol, hostname, and port) as the page where the request originates.

PR #191, opened 2025-03-06
Update docs for qps value

herrewig:

Docs should reflect new values we introduce

# What's changing?

Add documentation of new qps value


herrewig:
@dspigarelli and @Rickyxstar , confirming the collector doesn't use this env var?

dspigarelli:
> @dspigarelli and @Rickyxstar , confirming the collector doesn't use this env var? (it doesn't appear to be there)

The collector doesn't because it does all the querying via the service layer.

dspigarelli (approved):

herrewig:
> The collector doesn't because it does all the querying via the service layer.

Delicious!

PR #197, opened 2025-03-17
Allow assigning workloads to nodes

herrewig:

Users need the ability to assign workloads to nodes with specific labels. This can be done multiple ways but one of the [simplest and most common ways is via node labels](https://kubernetes.io/docs/tasks/configure-pod-container/assign-pods-nodes/). Our helm chart should provide the ability to run the Thoras platform on nodes with a specific label
 
# What's changing?

* Provide users the ability to run Thoras on nodes with a particular label
* Unit tests
* Updated docs

### DIFF
--- a/charts/thoras/tests/deployments_test.yaml
+++ b/charts/thoras/tests/deployments_test.yaml
@@ -1,9 +1,7 @@
 suite: Deployments
 templates:
   - api-server-v2/deployment.yaml
-  # - collector/deployment.yaml
   - dashboard/deployment.yaml
-  # - monitor/deployment.yaml
   - operator/deployment.yaml
   - reasoning-api/deployment.yaml
 set:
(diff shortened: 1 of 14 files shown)


herrewig:
FYI @parka1226 this is a totally different thing than what we talked about earlier. I focused on this next for timing purposes. Gonna work on the SELinux thing now but that will be a separate PR!

charts/thoras/tests/deployments_test.yaml
```diff
@@ -1,9 +1,7 @@
 suite: Deployments
 templates:
   - api-server-v2/deployment.yaml
-  # - collector/deployment.yaml
```
dspigarelli, line 4: Just double checking that we wanted to turn these deployment tests off.

dspigarelli (approved):
Changes look good to me.

PR #198, opened 2025-03-17
parent 0c4e6562bd10fd626e8a38aadfa7fd8e7c2f71ee

herrewig:

Sometimes users want to add additional `securityContext` objects to pod spec. In the most recent instance, we have a use-case where we want to do this for workloads that attach to PVs, but not for everything.

I've filed a ticket to do a more holistic configurability of securityContext for all workloads, but for now this is smallest useful thing

# What's changing?

* Allow providing additional security context settings for `collector` and `service-api-v2` -- the two workloads that mount a shared PV
* Unit tests
* Docs


charts/thoras/README.md
```diff
@@ -105,21 +105,23 @@ helm install \
 | metricsCollector.slackErrorsEnabled                             | Boolean | false            | Determines if error-level logs are sent to `slackWebHookUrl` |
 | metricsCollector.purge.logLevel                                 | String  | Nil              | Logging level                                                |
 | metricsCollector.init.imageTag                                  | String  | latest           | Image tag for metrics collector init container               |
+| metricsCollector.additionalPvSecurityContext | object | {} |                                            | Allows assigning additional securityContext objects to workloads that use PVs |
```
dspigarelli, line 108: [dust] Formatting could probably align better with the other rows in the md table.

dspigarelli (approved):

charts/thoras/README.md
herrewig: Good call! My ide font was too big to tell. Fixed!

parka1226 (approved):

dylan-moradpour (approved):

PR #201, opened 2025-03-19
Add configurable persistence accessMode for the PVC

dspigarelli:

Customers should be able to set an access mode to something other than `ReadWriteOnce` if they want to decouple the service layer from the timescale/collector pod.

### DIFF
--- a/charts/thoras/templates/collector/pvc.yaml
+++ b/charts/thoras/templates/collector/pvc.yaml
@@ -16,7 +16,7 @@ spec:
   {{- end }}
   storageClassName: {{ .Values.metricsCollector.persistence.storageClassName }}
   accessModes:
-    - {{ .Values.metricsCollector.persistence.accessMode | default "ReadWriteOnce" | quote }}
+    - {{ .Values.metricsCollector.persistence.accessMode | default "ReadWriteOnce" }}
   resources:
     requests:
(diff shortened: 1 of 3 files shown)


charts/thoras/templates/collector/pvc.yaml
```diff
@@ -16,7 +16,7 @@ spec:
   {{- end }}
   storageClassName: {{ .Values.metricsCollector.persistence.storageClassName }}
   accessModes:
-    - ReadWriteOnce
+    - {{ .Values.metricsCollector.persistence.accessMode | default "ReadWriteOnce" | quote }}
```
dspigarelli, line 19: I chose to make this only a single value instead of a list but I can make it a list if that's desired. Also, not sure if the `quote` is strictly necessary.

herrewig: > Also, not sure if the quote is strictly necessary.

Not necessary but a widely used convention, keep it!

herrewig: Also I'm +1 to passing a string rn instead of a list. Shipping the smallest useful thing

dspigarelli: I'm realizing adding `quote` is probably wrong since it results in this:
Screenshot 2025-03-21 at 9 30 17 AM

PR #203, opened 2025-03-19
Make accessmode a var

Rickyxstar:

Thoras users will need to be able to config their pvc acessmode

# What's changing?

added `.Values.metricsCollector.persistence.accessMode` and default to ReadWriteOnce


dspigarelli:
Closing in favor of https://github.com/thoras-ai/helm-charts/pull/201

PR #207, opened 2025-03-21
Update docs to reflect current thorasVersion

herrewig:

Docs should be aligned w/default Thoras version

# What's changing?

Update README w/correct thoras version


dspigarelli (approved):
:shipit:

PR #208, opened 2025-03-21
Add blob api service

dspigarelli:

In order to unbind the API service layer from the PVC and the collector deployment, we're adding a micro service that handles storage of blobs that will collocate inside the collector pod.

### DIFF
--- a/charts/thoras/templates/collector/deployment.yaml
+++ b/charts/thoras/templates/collector/deployment.yaml
@@ -101,7 +101,7 @@ spec:
         name: timescaledb
         imagePullPolicy: "{{ .Values.imagePullPolicy }}"
         ports:
-        - containerPort: 5432
+        - containerPort: {{ .Values.metricsCollector.timescale.containerPort }}
         env:
           - name: "POSTGRES_PASSWORD"
             valueFrom:
--- a/charts/thoras/templates/collector/deployment.yaml
+++ b/charts/thoras/templates/collector/deployment.yaml
@@ -37,7 +37,10 @@ spec:
         args:
           - |
             mkdir -p /var/lib/share/elasticsearch/data
+
+            # migrate non-elastic data to the new location
             find /var/lib/share/elasticsearch -not -name postgresql -not -name data -not -name blobs -mindepth 1 -maxdepth 1 -exec mv {} /var/lib/share/elasticsearch/data \;
+
             chown -R 1000:0 /var/lib/share/elasticsearch/data
             rm -f /var/lib/share/elasticsearch/data/node.lock
         {{- with .Values.metricsCollector.additionalPvSecurityContext }}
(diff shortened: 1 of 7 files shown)


charts/thoras/templates/collector/deployment.yaml
```diff
@@ -37,7 +37,10 @@ spec:
         args:
           - |
             mkdir -p /var/lib/share/elasticsearch/data
+
+            # migrate non-elastic data to the new location
```
dspigarelli, line 41: Adding spacing and comments for clarity.

charts/thoras/templates/collector/deployment.yaml
```diff
@@ -101,7 +109,7 @@ spec:
         name: timescaledb
         imagePullPolicy: "{{ .Values.imagePullPolicy }}"
         ports:
-        - containerPort: 5432
+        - containerPort: {{ .Values.metricsCollector.timescale.containerPort }}
```
dspigarelli, line 112: Not sure why we weren't using the config before but instead the hard coded value.

charts/thoras/templates/collector/service.yaml
```diff
@@ -15,7 +15,7 @@ spec:
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ .Values.metricsCollector.timescale.name }}
+  name: timescale
```
dspigarelli, line 18: No reason to parameterize this.

charts/thoras/tests/deployments_test.yaml
```diff
@@ -14,8 +14,6 @@ set:
     replicas: 12
   thorasDashboard:
     replicas: 12
-  thorasApiServer:
```
dspigarelli, line 17: Cleaning up legacy settings per [the Boy Scout rule](https://deviq.com/principles/boy-scout-rule)

charts/thoras/templates/collector/deployment.yaml
```diff
@@ -201,6 +209,38 @@ spec:
                 key: password
           - name: SERVICE_COLLECTION_PARALLELIZATION
             value: "{{ .Values.metricsCollector.collector.parallelization | default 20}}"
+      - image: {{ .Values.imageCredentials.registry }}/services:{{ default .Values.thorasVersion .Values.metricsCollector.collector.imageTag }}
+        imagePullPolicy: "{{ .Values.imagePullPolicy }}"
+        name: thoras-blob-api
+        command: ["/app/blob-service", "-p", "80"]
```
herrewig, line 215: Can we run this on 8080 to reduce one extra step when in the future we pay off the running-pods-as-root tech debt? I'm confident we'll do that at some point

charts/thoras/values.yaml
```diff
@@ -71,20 +71,10 @@ metricsCollector:
     imageTag: "2.17.2-pg16"
     name: timescale
     containerPort: 5432
+  blobService:
+    containerPort: 80
```
herrewig, line 75: [somewhere-between-pebble-and-boulder] See [this comment](https://github.com/thoras-ai/helm-charts/pull/208/files#r2008016296)

It's great if the service is 80 but we should have have containers themselves listen on unprivileged ports

charts/thoras/templates/collector/deployment.yaml
dspigarelli: I've updated the `containerPort` to 8080.

herrewig (approved):
LGTM!!

PR #212, opened 2025-03-24
Allow thoras op to scale any resource

Rickyxstar:

Thoras clients would like for Thoras to scale workloads that are not deployments.

# What's changing?

Updated thoras operator service account RBAC to allow it to get/update any scale subresource (rollouts, statefulsets)


herrewig (approved):
LGTM

PR #213, opened 2025-03-24
Ensure resource specs on all deployments

herrewig:

Ensuring all containers come with sane defaults for resource specs and the ability to override them is a basic chart feature we should already have. This gives users the ability to trust k8s will provision the necessary compute for components as well the ability to customize to their own needs

# What's changing?

* Add resource spec to deployments missing them
* Add unit tests that ensure all deployments have resource spec

### DIFF
--- a/charts/thoras/tests/all_deployments_test.yaml
+++ b/charts/thoras/tests/all_deployments_test.yaml
@@ -1,0 +1,27 @@
+suite: All Deployments
+templates:
+  - "*/deployment.yaml"
+set:
+  # Enable all deployments
+  thorasForecast:
+    worker:
+      enabled: true
+  thorasMonitor:
+    enabled: true
+    unittesting: true
+  thorasMonitorV2:
+    enabled: true
+    unittesting: true
+  thorasReasoning:
+    enabled: true
+tests:
+  - it: Object type correct
+    asserts:
+      - isKind:
+          of: Deployment
+  - it: Ensure resource requests and limits set
+    asserts:
+      - exists:
+          path: spec.template.spec.containers[*].resources.limits
+      - exists:
+          path: spec.template.spec.containers[*].resources.requests
--- a/charts/thoras/values.yaml
+++ b/charts/thoras/values.yaml
@@ -58,22 +58,47 @@ metricsCollector:
   podAnnotations: {}
   collector:
     name: thoras-collector
+    limits:
+      memory: "8192Mi"
+    requests:
+      cpu: "16m"
+      memory: "32Mi"
   search:
     imageTag: "8.12.1"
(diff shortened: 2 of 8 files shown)


charts/thoras/tests/all_deployments_test.yaml
```diff
@@ -0,0 +1,27 @@
+suite: All Deployments
```
herrewig, line 1: I felt like a unit test that covers _all_ deployments would be useful, so I created this new file. I moved the things that cover many-but-not-all deployments to select_deployments file

charts/thoras/values.yaml
```diff
@@ -58,22 +58,47 @@ metricsCollector:
   podAnnotations: {}
   collector:
     name: thoras-collector
+    limits:
+      memory: "8192Mi"
```
herrewig, line 62: I chose 8GB as an extremely-liberal-but-still-present ceiling as a memory limit default for all components that didn't currently have an intentionally defined limit

charts/thoras/values.yaml
```diff
@@ -119,11 +144,23 @@ thorasDashboard:
 
 thorasMonitor:
   enabled: false
+  unittesting: false
```
herrewig, line 147: This is a workaround to the fact the monitor deployment loads the contents of configmap into the pod spec which makes unittesting impossible. With this flag we can turn that functionality off -- allowing us to "skip" the line that does that

charts/thoras/values.yaml
```diff
@@ -58,22 +58,47 @@ metricsCollector:
...
+    limits:
+      memory: "8192Mi"
+    requests:
+      cpu: "16m"
+      memory: "128Mi"
   blobService:
     containerPort: 8080
     port: 80
+    limits:
+      memory: "8192Mi"
```
dspigarelli, line 98: Definitely doesn't need 8gb, but I'm not sure what the limit is.

herrewig: The limit just OOMs the pod if usage surpasses it. The fact we'd never need 8gb is great -- it's just a way to have _some_ ceiling in the event of an unknown state like a freak memory leak. We don't ever expect to hit this

charts/thoras/values.yaml
dspigarelli: Does this mean that the `metrics-collector` pod will need 4 x 8gb, or 32gb? Personally, that seems like a lot. If we need to cut space, I'd recommend that the blob server and the `metrics-collector` containers could probably run leaner, say 2gb.

dspigarelli: We could probably also work on extracting the collector into it's own deployment.

PR #214, opened 2025-03-24
4.23.0

herrewig:
Release 4.23.0


dspigarelli (approved):
I'd also be fine with 4.22.2, but numbers are cheap.

herrewig:
> I'd also be fine with 4.22.2, but numbers are cheap.

Hotfix works for me!

PR #215, opened 2025-03-24
Combine fix-dir-ownership init container

dspigarelli:

We should combine the `fix-dir-ownership` and `fix-dir-ownership-postgresql` into the same init container to both fix a helm chart issue that occurs when installing the 4.22.0 chart and also for readability.


herrewig (approved):
🍐 w/David, w00t!

PR #216, opened 2025-03-25
Propagate node tolerations to cleanup job

herrewig:

Users expect node tolerations to propagate to all workloads, even the metric cleanup job. Let's make sure it does

# What's changing?

* Propagate node tolerations to the metric cleanup job
* Add unit test that ensures every deployment always gets tolerations
* Move nodeSelector test to the "all deployments" test


Rickyxstar (approved):
LGTM! ⚡

PR #219, opened 2025-03-25
Pass the THORAS_NS env var to the service layer

dspigarelli:

We recently started using the `THORAS_NS` value in the service layer and need to make sure it stays updated.


herrewig (approved):
LGTM

PR #226, opened 2025-03-28
Make enqueue forecast setting a string

dylan-moradpour:

This env var needs to be a string, not a bool.


herrewig (approved):
LGTM!

PR #231, opened 2025-04-08
Upgrade Elastic and Timescale versions

dspigarelli:

Keep on the latest/greatest to address possible security vulnerabilities.

# What's changing?

Updating timescale and elastic images to the latest tags.


herrewig (approved):
LGTM! Thank you David!

PR #234, opened 2025-04-14
Deploy v2 monitor as another container in thoras-monitor

dspigarelli:

We plan on removing the v1 monitor completely. Instead of deploying it as yet-another-service during the transition period, I'm suggesting that we instead deploy it as another container in the existing `thoras-monitor` deployment. Then when we remove the deprecated v1 monitor, it will be as simple as removing the v1 monitor container.


charts/thoras/values.yaml
```diff
@@ -157,12 +157,6 @@ thorasMonitor:
 
 thorasMonitorV2:
   enabled: false
-  unittesting: false
```
dspigarelli, line 160: I'm advocating we adopt the same memory/cpu limits as the v1 monitor

Rickyxstar (approved):
LGMT 🚢

PR #242, opened 2025-04-24
Export api service layer prometheus scrape endpoint

dspigarelli:

We need to tell prometheus where to scrape our API service metrics.

# What's changing?

Adding annotation for the API service layer prometheus endpoint.


herrewig (approved):
LGTM!!

PR #244, opened 2025-04-28
Move prometheus-scrape-annotations

dspigarelli:

The prometheus scrape point annotations should be available regardless of other annotations.


herrewig (approved):
LGTM -- might want to make this [toggle-able in the future](https://github.com/cert-manager/cert-manager/blob/master/deploy/charts/cert-manager/templates/cainjector-deployment.yaml#L47-L54) but I think leaving it on as default works for now

dspigarelli:
> LGTM -- might want to make this [toggle-able in the future](https://github.com/cert-manager/cert-manager/blob/master/deploy/charts/cert-manager/templates/cainjector-deployment.yaml#L47-L54) but I think leaving it on as default works for now

Updated to add a `thorasApiServerV2.prometheus.enabled` flag that is default `true`

PR #246, opened 2025-04-29
Add support for service-specific labeling

dspigarelli:

Customers would like to specify per-service labels to drive monitoring, etc.

### DIFF
--- a/charts/thoras/templates/collector/deployment.yaml
+++ b/charts/thoras/templates/collector/deployment.yaml
@@ -2,6 +2,11 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
+  labels:
+    app.kubernetes.io/name: {{ .Chart.Name }}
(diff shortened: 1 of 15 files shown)


charts/thoras/templates/collector/deployment.yaml
```diff
@@ -2,6 +2,11 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
+  labels:
```
herrewig, line 5: Did we do this because we'll need to rebase against https://github.com/thoras-ai/helm-charts/pull/246/ after merge?

herrewig (approved):
LGTM!

charts/thoras/templates/collector/deployment.yaml
dspigarelli: I did this because when I was working on #246 I noticed it was missing. But yeah, #247 will be affected.

PR #247, opened 2025-04-29
Add common label template

dspigarelli:

When adding the ability to specify arbitrary labels to services and pods, I noticed that we could do a bit of DRYing and also found areas where we were missing the expected labels.

Note: I did not apply labels to the secrets, unsure of whether this would trigger an update to the secrets or not.


charts/thoras/templates/_helpers.tpl
```diff
@@ -1,3 +1,13 @@
 {{- define "imagePullSecret" }}
 {{- printf "{\"auths\": {\"%s\": {\"auth\": \"%s\"}}}" .Values.imageCredentials.registry (printf "%s:%s" .Values.imageCredentials.username .Values.imageCredentials.password | b64enc) | b64enc }}
 {{- end }}
+
```
dspigarelli, line 4: Hat tip to https://github.com/cert-manager/cert-manager/blob/06cd2fe0a2ff3e58fa8f017f09d4f7b7578f8810/deploy/charts/cert-manager/templates/_helpers.tpl#L150-L162

herrewig (approved):
LGTM

And regarding

> I did not apply labels to the secrets, unsure of whether this would trigger an update to the secrets or not.

I believe you're ok to do this, though I'm fine with leaving them off for now as I'm not 100% sure

herrewig (approved):
LGTM!

PR #249, opened 2025-04-30
Expose prometheus metric exporter on operator

dspigarelli:

We need to register the prometheus exporter on the operator so that the prometheus server starts to ingest metrics from it.


parka1226 (approved):
LGTM!

PR #251, opened 2025-04-30
Enable v2 monitor by default

dspigarelli:

We should default enable the v2 monitor as we begin migrating functionality from the v1 monitor


herrewig (approved):
LGTM

PR #252, opened 2025-04-30
Add vertical limit config to CRD

Rickyxstar:

Thoras user will like the ability to set how much Thoras mutates their limits to keep them inline with their requests. 

# What's changing?

Made upper bound omitable and added optional limit prop. 

```
vertical:
  containers:
    - name: foo
      memory:
        lowerbound: 128Mi
        limit:
          ratio: 1.5
```


dspigarelli (approved):
:shipit:

PR #253, opened 2025-05-01
Add name to service ports

dspigarelli:

Add `name` to service ports to allow service monitor targeting.

### DIFF
--- a/charts/thoras/templates/api-server-v2/service.yaml
+++ b/charts/thoras/templates/api-server-v2/service.yaml
@@ -10,7 +10,8 @@ metadata:
 {{- end }}
 spec:
   ports:
-  - port: {{ .Values.thorasApiServerV2.port }}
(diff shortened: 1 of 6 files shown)


charts/thoras/templates/api-server-v2/service.yaml
```diff
@@ -10,7 +10,8 @@ metadata:
 {{- end }}
 spec:
   ports:
-  - port: {{ .Values.thorasApiServerV2.port }}
+  - name: thoras-api-server-v2
```
herrewig, line 13: Not necessarily a blocker but my understanding of the convention of port names on a service object is to describe the protocol more than application-specific things. Like name would be `https` if the port is 443 or `http` for `8080`, `grpc`, etc

herrewig (approved):
🔥

PR #259, opened 2025-05-06
Expose prometheus metrics port on operator

dspigarelli:

The prometheus metrics service on the operator should be exposed so service monitors can target it.


herrewig (approved):
[dust / sand] for TLS connections I would name it https over http but definitely not a blocker, your call

LGTM!

PR #261, opened 2025-05-08
Update third-party image tags

dspigarelli:

Updating third-party images to address various vulnerabilities


herrewig (approved):
LGTM

PR #264, opened 2025-05-16
Remove forecastQueue.enabled flags

dspigarelli:

All forecasts are run by the queue, not individual cron jobs


Rickyxstar (approved):
LGTM! ⚡

PR #265, opened 2025-05-20
Ensure empty string for slack webhook URL

herrewig:

We want to make sure there's an empty string as the value of this secret no matter what. Also this is best practice

# What's changing?

Wrap value in quotes

### DIFF
--- a/charts/thoras/templates/monitor/secret.yaml
+++ b/charts/thoras/templates/monitor/secret.yaml
@@ -2,7 +2,7 @@
 ---
 apiVersion: v1
 data:
-  webhookUrl: {{ .Values.slackWebhookUrl | b64enc }}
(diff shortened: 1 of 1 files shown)


charts/thoras/templates/monitor/secret.yaml
```diff
@@ -2,7 +2,7 @@
 ---
 apiVersion: v1
 data:
-  webhookUrl: {{ .Values.slackWebhookUrl | b64enc }}
+  webhookUrl: "{{ .Values.slackWebhookUrl | b64enc }}"
```
dspigarelli, line 5: [dust] Could also do:
```suggestion
  webhookUrl: {{ .Values.slackWebhookUrl | b64enc | quote }}
```

dspigarelli (approved):
:shipit:

herrewig:
I like it!

PR #266, opened 2025-05-20
Remove unused collector environment variable

dspigarelli:

The collector no longer talks directly to postgres and doesn't need the DSN.


parka1226 (approved):
LGTM! Just a quick question, do we want to keep postgres dns variable inside of charts/thoras/templates/collector/cronjob.yaml?

dspigarelli:
> LGTM! Just a quick question, do we want to keep postgres dns variable inside of charts/thoras/templates/collector/cronjob.yaml?

Great catch! Yes, we want to remove that one too.

PR #268, opened 2025-05-21
Added v2 dashboard

Rickyxstar:
## Purpose of the Change

We are implementing this change to enable Thoras users to experience our newly enhanced dashboard.

## Changes Implemented

The V2 Thoras dashboard will be deployed by default. Users can revert to the V1 dashboard by modifying the following configuration:
```
thorasDashboard:
  v2:
    enabled: false
```

### DIFF
--- a/charts/thoras/tests/__snapshot__/dashboard_configmap_test.yaml.snap
+++ b/charts/thoras/tests/__snapshot__/dashboard_configmap_test.yaml.snap
@@ -1,0 +1,132 @@
+Should default to v2 port:
+  1: |
+    apiVersion: v1
+    data:
+      dashboard-v2.nginx.conf: |
+        events {
+          worker_connections 1024;
+        }
+
+        http {
+          include       mime.types;
+
+          server {
+            listen       5173;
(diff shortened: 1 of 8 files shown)


charts/thoras/tests/__snapshot__/dashboard_configmap_test.yaml.snap
```diff
@@ -0,0 +1,132 @@
+Should default to v2 port:
```
herrewig, line 1: ❤️

charts/thoras/values.yaml
```diff
@@ -135,6 +135,9 @@ thorasApiServerV2:
 
 thorasDashboard:
   enabled: true
+  v2:
+    enabled: true
```
herrewig, line 139: I assume we just set this to false and the ternary will flip nginx?

herrewig (approved):
Amazing!

charts/thoras/values.yaml
Rickyxstar: That's correct!

PR #271, opened 2025-05-30
Update AiScaleTarget with previousVerticalMode status field

dspigarelli:

We recently made changes to our AiScaleTarget CRD that need to be reflected in the helm chart.


charts/thoras/templates/crd/aiscaletarget.yaml
```diff
@@ -5,8 +5,6 @@ metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.17.2
   name: aiscaletargets.thoras.ai
-  labels:
```
dspigarelli, line 8: Dropping the labels portion because we don't expect this to be used and only adds complexity.

PR #275, opened 2025-06-02
Add SERVICE_POD_LOG_LABEL

garrettporter:
SERVICE_POD_LOG_LABEL is used to dynamically retrieve Thoras service names.

# What's changing?
Added environment variable SERVICE_POD_LOG_LABEL which defaults to "app.kubernetes.io/name=thoras"


herrewig (approved):
LGTM , paired with @garrettporter ! Let's goooo!!!

dspigarelli (approved):
:shipit:

PR #277, opened 2025-06-03
Give operator role the ability to restart/patch deployments

dspigarelli:

The operator recently started restarting deployments/etc. when a VPA target moved from `autonomous` to `recommendation`. This is done in the reconciler but it didn't have the correct permissions.


herrewig (approved):
LGTM

PR #282, opened 2025-06-13
Make Model Mode optional

garrettporter:
`model.mode` is no longer a required field. See https://github.com/thoras-ai/platform/pull/1473

# What's changing?
Removed required tag and updated description.


herrewig (approved):
LGTM!

PR #284, opened 2025-06-18
Give operator ability to patch more than just deployments

dspigarelli:

We support workloads other than `deployments`, including `stateful` sets. The operator and API service layer both should have privileges to `patch`, necessary for triggering workload restarts.

# What's changing?

- opening up the RBAC roles for the operator and the API service layer to have permission to patch `*`, not just `deployments`
- removing the dashboard's ability to patch deployments since this is no longer necessary

### DIFF
--- a/charts/thoras/templates/api-server-v2/rbac.yaml
+++ b/charts/thoras/templates/api-server-v2/rbac.yaml
@@ -51,6 +51,7 @@ rules:
   - '*'
 - apiGroups:
   - apps
+  - argoproj.io
   resources:
   - '*'
   verbs:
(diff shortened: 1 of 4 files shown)


charts/thoras/templates/api-server-v2/rbac.yaml
```diff
@@ -52,7 +52,7 @@ rules:
 - apiGroups:
   - apps
```
Rickyxstar, line 53: This should be a wild card too. Argo Rollouts live in the `argoproj.io` api group
```suggestion
  - '*'
```

Rickyxstar (approved):
Left a comment! Other than that looks good to me!

charts/thoras/templates/api-server-v2/rbac.yaml
dspigarelli: I updated this to add `argoproj.io` to the list of `apiGroups` and tested it out locally to make sure it worked.

PR #288, opened 2025-06-23
Add readiness probe to API service

dspigarelli:

We have a `/health` endpoint on the service layer and just needed to add a `readinessProbe` section to use it.


charts/thoras/templates/api-server-v2/deployment.yaml
```diff
@@ -118,6 +118,10 @@ spec:
                 fieldPath: metadata.namespace
         ports:
         - containerPort: {{ .Values.thorasApiServerV2.containerPort }}
+        readinessProbe:
```
herrewig, line 121: What is the health endpoint doing? Is it making a round-trip ping to the DB? that would be ideal for a readiness probe. Anything less should probably be a liveness probe

herrewig (approved):
LGTM!

PR #303, opened 2025-07-22
Pass ThorasVersion to collector as env var

dspigarelli:

By passing the `ThorasVersion` as an environment variable to the collector, the migrator can take steps to backup or restore the database, if necessary.

# What's changing?

Passing the `ThorasVersion` (e.g. 4.32.0) to the collector as `SERVICE_THORAS_VERSION`.


herrewig (approved):
LGTM!

PR #307, opened 2025-07-29
Update timescale password on collector start

dspigarelli:

In certain scenarios the timescale secret is being changed. For an existing deployment, this cascades to services being unable to communicate with the database until the database user's password is updated.

# What's changing?

When the `metrics-collector` starts up, we wait for timescale/postgres to be available, then sync the `postgres` user password.

### DIFF
--- a/charts/thoras/templates/collector/deployment.yaml
+++ b/charts/thoras/templates/collector/deployment.yaml
@@ -147,6 +147,7 @@ spec:
         command: ["/bin/sh", "-c"]
         args:
           - |
+            # wait for elasticsearch to be available
             retries=0
             max_retries=48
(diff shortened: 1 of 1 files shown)


Rickyxstar:
Should we do the same thing for Elastic?

dspigarelli:
> Should we do the same thing for Elastic?

No, it's not necessary since elastic will use whatever `ELASTIC_PASSWORD` is set to. Postgres instead uses `POSTGRES_PASSWORD` only during initialization of the database but any changes after that need to have the database user updated. I tested it out by changing the `thoras-elastic-password` secrets and restarting the `metrics-collector` and elastic was immediately accessible with the new password

PR #309, opened 2025-08-05
Add namespaced role bindings

dspigarelli:

Per customer request, our helm chart's RBAC can now be scoped to specific namespaces via `Role` and `RoleBinding` resources instead of `ClusterRole` and `ClusterRoleBinding`. Default behavior is `ClusterRole`/etc.

# What's changing?

Adding a new value, `rbac.namespaces` which defaults to an empty list, but which can be a list of namespace `string` values for which individual `Role` and `RoleBinding` objects will be created.


charts/thoras/templates/collector/rbac.yaml
```diff
@@ -1,45 +1,133 @@
...
+kind: Role
 metadata:
-  labels:
-    {{- include "labels" . | nindent 4 }}
-  name: thoras-collector
-  namespace: {{ .Release.Namespace }}
-imagePullSecrets:
-{{- if .Values.imageCredentials.secretRef }}
-  - name: {{ .Values.imageCredentials.secretRef }}
+  name: {{ print "thoras-collector-" .rbacNamespace }}
```
dspigarelli, line 7: Future PR to clean this up so the service account can be renamed like we do for other services accounts

Rickyxstar (approved):

PR #311, opened 2025-08-06
Remove v1 dashboard

dspigarelli:

We no longer use the v1 dashboard and should remove it from deployments.

# What's changing?

Removing the v1 dashboard from the chart.


Rickyxstar (approved):
⚡

PR #312, opened 2025-08-06
Trim unnecessary rbac

dspigarelli:

We should remove permissions we previously needed but don't need any longer.

# What's changing?

Dropping RBAC roles/etc. for dashboard, et al.


charts/thoras/templates/reasoning-api/deployment.yaml
```diff
@@ -25,7 +25,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
     spec:
-      serviceAccountName: thoras-collector
+      serviceAccountName: {{ .Values.thorasReasoning.serviceAccount.name }}
```
dspigarelli, line 28: The reasoning component was previously piggy backing on the collector's service account.

PR #313, opened 2025-08-07
Don't use the fully qualified service URL, just the intra-namespace URL

dspigarelli:

The init containers don't need to reference the fully qualified URL for the service but can instead use the intra-namespace URL.


herrewig (approved):
LGTM!

PR #318, opened 2025-08-11
Give the API service up to 5 minutes to start

dspigarelli:

When a cluster has a large number of ASTs, it can take a couple minutes to load the initial catalog from k8s. Adding a [startupProbe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes) gives the API service long enough to load the catalog.

# What's changing?


herrewig (approved):
LGTM!!

PR #321, opened 2025-08-13
Feat/retrain steady state config

muditmathur11:

We’ve seen better adaptation to new/shifted traffic patterns when models are retrained more frequently in steady state. And, was not possible to enable this change without updating/deploying Python code. 

Closes https://github.com/thoras-ai/platform/issues/1766

# What's changing?

- In `charts/thoras/templates/forecast-worker/deployment.yaml`, optionally inject `RETRAIN_STEADY_STATE_HOURS` into worker container when `values.yaml` sets `thorasForecast.worker.retrainSteadyStateHours. 
    - We clamp to >= 1.0, and render as a string 
    - (NOTE: In the corresponding Python codebase updates, we use a steady-state default value of 12 hours if this value is not set) 

# Testing
Added the following unit tests:
- Integer and float inputs render (e.g., 10 → "10").
- null / "" omit the env var.
- Values below 1.0 are clamped to "1".

Also checked this all manually with `helm template ...`

### DIFF
--- a/charts/thoras/tests/forecast_worker_test.yaml
+++ b/charts/thoras/tests/forecast_worker_test.yaml
@@ -83,7 +83,7 @@ tests:
           content:
             name: RETRAIN_STEADY_STATE_HOURS
             value: "10"
-            
+
   - it: omits RETRAIN_STEADY_STATE_HOURS when value is null
     set:
       thorasForecast:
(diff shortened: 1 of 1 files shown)


herrewig (approved):
Amazing! Way to learn this so fast!!

PR #327, opened 2025-08-15
Update third-party images

dspigarelli:

Updating our third-party images keeps on top of security updates.

# What's changing?

Updating the elastic and timescale image tags.


herrewig (approved):
LGTM

PR #335, opened 2025-08-27
Add test for worker service account

dspigarelli:

Consistently tested code, include whether a service account was created correctly, reduces risk of incorrect code.

# What's changing?

Adding a unit test for the worker receiving the correct product key.


herrewig (approved):
🚢

PR #358, opened 2025-09-16
Enable thoras worker by default

dspigarelli:
## How does this help customers?

Customers get the worker component enabled by default, reducing configuration overhead for new deployments.

## What's changing?

- Changed `thorasWorker.enabled` default from `false` to `true` in values.yaml
- Updated README.md documentation to reflect new default
- Removed explicit `thorasWorker.enabled: true` from test files since it's now the default

## How Tested

Updated all existing helm unit tests to work with the new default. Tests pass without requiring explicit worker enablement.

## Claude Prompts

User: "Create a PR for the current branch using --head, tag Mudit for review. Be super concise in PR description."

/cc @muditmathur11

### DIFF
--- a/charts/thoras/README.md
+++ b/charts/thoras/README.md
@@ -148,7 +148,7 @@ helm install \
 
 | Key                              | Type    | Default       | Description                                                  |
 | -------------------------------- | ------- | ------------- | ------------------------------------------------------------ |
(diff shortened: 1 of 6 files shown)


herrewig (approved):
LGTM

PR #374, opened 2025-10-02
Increase timescale default requests

dspigarelli:

With elastic no longer necessary, we need to commit more resources to timescale to avoid OOMs.

```
POD                                 NAME               CPU(cores)   MEMORY(bytes)
metrics-collector-c88d8c796-945wr   thoras-blob-api    20m          158Mi
metrics-collector-c88d8c796-945wr   thoras-collector   0m           12Mi
metrics-collector-c88d8c796-945wr   timescaledb        692m         5012Mi
```

# What's changing?

Increasing timescale's request to 3gb

### DIFF
--- a/charts/thoras/values.yaml
+++ b/charts/thoras/values.yaml
@@ -91,7 +91,7 @@ metricsCollector:
       memory: "8192Mi"
     requests:
       cpu: "400m"
-      memory: "1280Mi"
+      memory: "3072Mi"
   blobService:
     pprof:
       enabled: false
--- a/charts/thoras/values.yaml
+++ b/charts/thoras/values.yaml
@@ -88,7 +88,7 @@ metricsCollector:
     name: timescale
     containerPort: 5432
     limits:
-      memory: "8192Mi"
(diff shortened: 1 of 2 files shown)


herrewig:
To prevent OOMs, we need to up the limit, not the request

charts/thoras/values.yaml
```diff
@@ -91,7 +91,7 @@ metricsCollector:
       memory: "8192Mi"
     requests:
       cpu: "400m"
-      memory: "1280Mi"
+      memory: "3072Mi"
```
herrewig, line 94:
```suggestion
      memory: "1280Mi"
```

charts/thoras/values.yaml
```diff
@@ -91,7 +91,7 @@ metricsCollector:
       memory: "8192Mi"
```
herrewig, line 91:
```suggestion
      memory: "16384Mi"
```

herrewig (approved):
LGTM!

PR #375, opened 2025-10-02
Added watch

Rickyxstar:

Gives the Thoras worker the ability to watch objects.

# What's changing?

Added watch to throas worker rbac

# Forget anything

[ x ] Do any feature flags need to be managed?


herrewig (approved):
LGTM

PR #384, opened 2025-10-06
Align SLM compute requests with inference need

herrewig:

SLM will literally fail to initialize if there's not enough memory for inference available on the system. No relying on spiking is allowed, so we'll need to pre-allocate what we need

# What's changing?

Explicitly allocate what we need for inference

# Forget anything

[x] Do any feature flags need to be managed?

### DIFF
--- a/charts/thoras/values.yaml
+++ b/charts/thoras/values.yaml
@@ -256,7 +256,7 @@ thorasSlmAgent:
     memory: 8192Mi
   requests:
     cpu: 1024m
-    memory: 1024Mi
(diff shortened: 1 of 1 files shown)


charts/thoras/values.yaml
```diff
@@ -256,7 +256,7 @@ thorasSlmAgent:
     memory: 8192Mi
   requests:
     cpu: 1024m
-    memory: 1024Mi
+    memory: 4096Mi
```
herrewig, line 259: 3.7 is required for inference

PR #386, opened 2025-10-07
Remove outdated collector parameter

dspigarelli:

To start, the collector must no longer use the `-c` parameter.

# What's changing?

Dropping the out-dated `-c` parameter. Future PR to remove the collector altogether.


herrewig (approved):
LGTM

PR #644, opened 2026-04-09
add suggestion fields to AST status

garrettporter:
# What's changing and why?
Track forecast errors and the most recent suggestion reconciled by the operator on the AST status. Reading directly from the AST status will allow the operator to make scaling decisions without needing to fetch from the suggestions table.


muditmathur11 (approved):
Nice!

PR #659, opened 2026-04-14
set enableDeploymentMonitor default to true

garrettporter:
# What's changing and why?
Use the new and improved deployment monitor by default.


dspigarelli (approved):

garrettporter:
I tested this by creating two pods in the Thoras namespace.

The first is stuck in the ImagePullBackoff state and does alert
`kubectl create deployment test-stuck --image=nginx:does-not-exist -n thoras`

The second is stuck in initializing and does not alert
```
kubectl apply -n thoras -f - <<EOF
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-initializing
spec:
  replicas: 1
  selector:
    matchLabels:
      app: test-initializing
  template:
    metadata:
      labels:
        app: test-initializing
    spec:
      initContainers:
      - name: init-sleep
        image: busybox
        command: ["sleep", "3600"]
      containers:
      - name: main
        image: nginx
EOF
```

PR #670, opened 2026-04-21
pass platform version to monitor

garrettporter:
Closes https://github.com/thoras-ai/platform/issues/4132

# What's changing and why?
Setting monitor thorasVersion, which is consistant with other services.


herrewig (approved):
LGTM

PR #676, opened 2026-04-27
Auto-size memory requests based on cluster state

dspigarelli:
## What's changing and why?

Adds `featureFlags.dynamicResourceSizing` (default off). When enabled, `api-server-v2`, `operator`, and `worker` memory requests are computed at install/upgrade time via Helm `lookup` from live pod and workload counts. Forecast-worker replicas also scale (`ceil(workloads / workloadsPerForecastWorker)`). Falls back to today's hardcoded values when lookup is unavailable.

## How tested

- All 226 helm unit tests pass, 50 snapshots unchanged
- `helm template` with flag off renders today's hardcoded values
- `helm template` with flag on + no kube context falls back correctly

### DIFF
--- a/charts/thoras/templates/_helpers.tpl
+++ b/charts/thoras/templates/_helpers.tpl
@@ -122,3 +122,117 @@ true
 {{ toYaml $out -}}
 {{- end -}}
 {{- end }}
+
+{{/*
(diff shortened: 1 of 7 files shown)


herrewig:
Spoke with @dspigarelli , we're going to hold off on merging this until we get more testing/mileage with it

dspigarelli:
Closing for now

PR #683, opened 2026-04-28
add operator readiness probe

garrettporter:

Wires up the operator's new `/readyz` endpoint to Kubernetes, so the operator pod is
removed from Service endpoints during restarts until its informer caches are synced and
the webhook server is ready. This prevents webhook timeouts and the resulting 409 conflicts
on scale operations that can occur when traffic is routed to an unready pod.

## What's changing

- Adds `readinessProbe` to the operator Deployment pointing at `/readyz` on port 8081
- Exposes port 8081 (`http-health`) on the operator container

Depends on platform change: [](https://github.com/thoras-ai/platform/pull/4224)

## How Tested

Deployed to minikube and confirmed pods are held unready during startup and only receive
webhook traffic once `/readyz` returns 200.

### DIFF
--- a/charts/thoras/templates/operator/deployment.yaml
+++ b/charts/thoras/templates/operator/deployment.yaml
@@ -215,8 +215,9 @@ spec:
             path: /readyz
             port: http-health
           initialDelaySeconds: 5
-          periodSeconds: 10
-          failureThreshold: 3
+          periodSeconds: 5
+          failureThreshold: 2
+          timeoutSeconds: 2
         {{- if .Values.thorasOperator.resources }}
         resources: {{ .Values.thorasOperator.resources | toYaml | nindent 10 }}
         {{- else }}
--- a/charts/thoras/templates/operator/deployment.yaml
+++ b/charts/thoras/templates/operator/deployment.yaml
@@ -210,11 +210,17 @@ spec:
             containerPort: {{ .Values.thorasOperator.prometheus.port }}
           - name: http-health
             containerPort: 8081
+        startupProbe:
+          httpGet:
+            path: /readyz
+            port: http-health
+          periodSeconds: 5
+          failureThreshold: 60
+          timeoutSeconds: 2
         readinessProbe:
           httpGet:
             path: /readyz
             port: http-health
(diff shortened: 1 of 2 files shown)


charts/thoras/templates/operator/deployment.yaml
```diff
@@ -208,6 +208,15 @@ spec:
...
           - name: http-metrics
             containerPort: {{ .Values.thorasOperator.prometheus.port }}
+          - name: http-health
+            containerPort: 8081
+        readinessProbe:
+          httpGet:
+            path: /readyz
+            port: http-health
+          initialDelaySeconds: 5
+          periodSeconds: 10
```
herrewig: does this mean it can be in a bad state for 30s? If so , I think that's too generous

What do you think about this:

```
    periodSeconds: 5
    failureThreshold: 2      # ~10s to detect
    timeoutSeconds: 2        # probe itself fails fast if server is hung
```

herrewig: Question @garrettporter @Rickyxstar : should we also add a startupProbe?

```
On customer cluster, informer cache sync at startup can take 60-120s (large pod/HPA/Deployment caches across all namespaces). controller-runtime doesn't open the webhook port until caches sync, so the readinessProbe will hit "connection refused" during that whole window and mark the pod NotReady. That's correct (pod genuinely isn't ready), but if we ever add a livenessProbe with similar tightness, kubelet will start restarting the pod mid-sync : death spiral. A startupProbe gates both:

  startupProbe:
    httpGet: { path: /readyz, port: 9443, scheme: HTTPS }
    periodSeconds: 5
    failureThreshold: 60   # 5 min cap on cold start
    timeoutSeconds: 2
```

garrettporter: Makes sense to me. I added
 ```
        startupProbe:
          httpGet:
            path: /readyz
            port: http-health
          periodSeconds: 5
          failureThreshold: 60
          timeoutSeconds: 2
```

herrewig (approved):
LGTM

PR #687, opened 2026-05-01
Remove forecast worker readiness feature flag

dspigarelli:
# What's changing and why?

Removes `enableForecastWorkerReadinessProbe` and `enableForecastWorkerLivenessProbe` feature flags — probes are now always enabled. Also removes the init container fallback that was used when the readiness probe was off.

## How Tested

Helm unit tests pass.

### DIFF
--- a/charts/thoras/templates/forecast-worker/deployment.yaml
+++ b/charts/thoras/templates/forecast-worker/deployment.yaml
@@ -84,10 +84,8 @@ spec:
           - name: METRICS_PORT
             value: {{ .Values.thorasForecast.prometheus.port | quote }}
           {{- end }}
-          {{- if .Values.featureFlags.enableForecastWorkerLivenessProbe }}
           - name: WORKER_HEARTBEAT_FILE
             value: /var/run/thoras/worker-heartbeat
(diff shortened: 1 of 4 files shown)


muditmathur11 (approved):

charts/thoras/templates/forecast-worker/deployment.yaml
```diff
@@ -128,21 +105,18 @@ spec:
...
           failureThreshold: 48
-        {{- end }}
-        {{- if .Values.featureFlags.enableForecastWorkerLivenessProbe }}
         livenessProbe:
           exec:
             command:
               - /bin/sh
               - -c
-              - "test $(( $(date +%s) - $(stat -c %Y /var/run/thoras/worker-heartbeat 2>/dev/null || echo 0) )) -lt 300"
+              - "test $(( $(date +%s) - $(stat -c %Y /var/run/thoras/worker-heartbeat 2>/dev/null || echo 0) )) -lt 600"
```
dspigarelli, line 111: Increase the window to 10 minutes since training can take longer than five minutes

PR #694, opened 2026-05-05
Add pprof enable flags for operator and worker

dspigarelli:
# What's changing and why?

Adds `thorasOperator.pprof.enabled` and `thorasWorker.pprof.enabled` flags (default: `false`), propagated as `SERVICE_ENABLE_PPROF` env vars. Allows enabling pprof endpoints per-component without a chart change.


muditmathur11 (approved):
LGTM

PR #695, opened 2026-05-05
Remove chunk interval dry run feature flag

dspigarelli:
# What's changing and why?

Removes the `featureFlags.enableChunkIntervalDryRun` flag and its `SERVICE_ENABLE_CHUNK_INTERVAL_DRY_RUN` env var from the worker — the feature is no longer gated.


muditmathur11 (approved):
LGTM

PR #700, opened 2026-05-06
Add health probes to forecast worker

dspigarelli:
# What's changing and why?

Adds startup, readiness, and liveness probes to the forecast worker using the existing `/health` endpoint on the metrics port.


charts/thoras/templates/worker/deployment.yaml
```diff
@@ -219,7 +219,7 @@ spec:
           - name: SERVICE_ENABLE_NODE_DETAILS_COLLECTOR
             value: {{ .Values.featureFlags.enableNodeDetailsCollector | ternary "true" "false" | quote}}
           - name: SERVICE_ENABLE_CHECK_DATABASE_HEALTH
-            value: {{ .Values.featureFlags.enableCheckDatabaseHealth | ternary "true" "false" | quote }}
+            value: "true"
```
dspigarelli, line 222: This needs to be `true` because the health endpoint uses database connectivity to decide if the worker is healthy or not.

Rickyxstar (approved):

PR #702, opened 2026-05-06
track OOMed containers

garrettporter:
# What's changing and why?
Tracks OOMed containers from the most recent collection cycle, so those specific containers can have memory scaled up.


herrewig (approved):
LGTM

PR #703, opened 2026-05-06
Increase baseline memory limits and requests

dspigarelli:
# What's changing and why?

Bumps memory limits from 2Gi → 4Gi and requests from 128–256Mi → 1Gi for the operator, API server, and worker. Aligns defaults with observed production usage to reduce OOMKills on fresh installs.


herrewig (approved):
LGTM

PR #708, opened 2026-05-07
feat: add percentile mode to AIScaleTarget and ClusterAIScaleTemplate…

muditmathur11:
… CRDs

# What's changing and why?

Syncs CRD schemas with [platform repo change](https://github.com/thoras-ai/platform/pull/4307) (pending merge).

AIScaleTarget and ClusterAIScaleTemplate CRDs:
- percentile added to mode enum on ModelSpec and MetricConfig
- percentile_value field added (x-kubernetes-int-or-string) to both structs at top-level and per-metric
- CEL rules added: percentile_value required when mode: percentile (top-level only), forbidden when mode is not percentile (both levels)

### DIFF
--- a/charts/thoras/templates/crd/aiscaletarget.yaml
+++ b/charts/thoras/templates/crd/aiscaletarget.yaml
@@ -689,8 +689,18 @@ spec:
                           - undershoot
                           - assurance_optimized
                           - overshoot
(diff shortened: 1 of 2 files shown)


herrewig (approved):
LGTM

PR #711, opened 2026-05-08
Add AFFINITY_ID env var to forecast worker

Rickyxstar:
## What's changing and why?

Sets `AFFINITY_ID` on the forecast worker container from the pod's `metadata.uid` via the downward API, giving each worker pod a unique, stable affinity identifier for the lifetime of the pod.

## How tested?

Helm unit tests pass; snapshot updated.


herrewig (approved):
LGTM

PR #712, opened 2026-05-08
refactor: remove Thoras AST resources and add scaling feature flags

Rickyxstar:
# What's changing and why?

Removes the obsolete `ast.yaml` templates across all components (api-server-v2, collector, dashboard, monitor, operator, worker) along with the related `external_timescale_test.yaml` test.

Adds new `featureFlags.thorasManageThoras` env vars to the worker deployment so Thoras can autoscale its own components, and refreshes the global label snapshot accordingly.

## How Tested

Helm unit tests pass; snapshot diff matches the new env vars added to the worker deployment.


dspigarelli (approved):
I'm approving but we should consider the impact to existing customers before we merge this and add it to the latest chart.

PR #715, opened 2026-05-11
Add readiness probe for API server

dspigarelli:
# What's changing and why?

Adds a readiness probe to the API server that checks `/health?db=true`, ensuring traffic is only routed once the DB connection is healthy. Also names the port (`http`) and tightens probe configs (startup threshold bumped to 30, liveness/readiness with explicit timeouts).


muditmathur11 (approved):
LGTM

PR #717, opened 2026-05-11
Add probes to monitor

dspigarelli:
# What's changing and why?

Replaces the init container (which blocked startup) with readiness/liveness probes on the monitor container that poll the API health endpoint. This aligns with how other components handle API dependency.


muditmathur11 (approved):
✅

PR #718, opened 2026-05-11
Add probes to timescale container

dspigarelli:
# What's changing and why?

Adds startup, readiness, and liveness probes to the timescale container using `pg_isready`.

Startup probe gives postgres up to 5 minutes to initialize; readiness and liveness probes detect connection failures and trigger restarts.

### DIFF
--- a/charts/thoras/templates/collector/deployment.yaml
+++ b/charts/thoras/templates/collector/deployment.yaml
@@ -165,6 +165,24 @@ spec:
           limits: {{ .Values.metricsCollector.timescale.limits | toYaml | nindent 12 }}
           requests: {{ .Values.metricsCollector.timescale.requests | toYaml | nindent 12 }}
         {{- end }}
+        startupProbe:
+          exec:
+            command: ["pg_isready", "-U", "postgres"]
+          periodSeconds: 5
+          failureThreshold: 60
+          timeoutSeconds: 2
+        readinessProbe:
+          exec:
+            command: ["pg_isready", "-U", "postgres"]
+          periodSeconds: 5
+          failureThreshold: 3
+          timeoutSeconds: 2
+        livenessProbe:
+          exec:
+            command: ["pg_isready", "-U", "postgres"]
+          periodSeconds: 10
+          failureThreshold: 6
+          timeoutSeconds: 2
         volumeMounts:
         {{- if .Values.metricsCollector.timescale.config.enabled }}
           - name: timescale-config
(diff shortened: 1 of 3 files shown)


muditmathur11 (approved):
LGTM

charts/thoras/templates/collector/deployment.yaml
```diff
@@ -165,6 +165,24 @@ spec:
...
+            command: ["pg_isready", "-U", "postgres"]
+          periodSeconds: 5
+          failureThreshold: 3
+          timeoutSeconds: 2
+        livenessProbe:
+          exec:
+            command: ["pg_isready", "-U", "postgres"]
+          periodSeconds: 10
+          failureThreshold: 6
+          timeoutSeconds: 2
```
herrewig: Is there ever a situation where postgres is slow to respond for a bit and we keep killing the container?

Maybe not, but I do worry In general about liveness being too strict, as a general thing

Not saying 2 second timeout is wrong

I don't want to sow doubt into this PR , I just want to make sure we're thinking through as much as possible and we're as confident as we can be 

If there were an area that would be more lenient , it would be here . Generally

I trust your judgment!

herrewig (approved):
LGTM

charts/thoras/templates/collector/deployment.yaml
dspigarelli: Good point. I'm going to up the time. Here are some additional thoughts:

> The setup is actually well-defended against the scenario you're describing. A few things to note:
> 
> The startup/liveness split is doing its job correctly. The startupProbe (line 168) runs first with failureThreshold: 60 × periodSeconds: 5 = 5 minutes of tolerance for slow initial startup. The liveness probe only activates after startup succeeds, so slow boot isn't a risk.
> 
> pg_isready is about as lightweight as it gets. It's a TCP connection check — it doesn't run a query. Postgres would have to be completely unable to accept connections (frozen, hung, OOM'd) for it to fail, not just slow under load. Query slowness won't trip this.
> 
> 60 seconds of grace. failureThreshold: 6 × periodSeconds: 10 means postgres needs to be entirely unresponsive for a full minute before a kill. At that point, a restart is probably the right call anyway.
> 
> One thing to scrutinize more: the readinessProbe (line 174) is stricter — failureThreshold: 3 × periodSeconds: 5 = 15 seconds before traffic is cut off. That's independent of liveness and won't kill the pod, but it could starve the blob service of its local postgres dependency quickly. That said, same argument applies — pg_isready failing for 15s means something's genuinely wrong.
> 
> The liveness config is about as lenient as I'd want it to be without risking a zombie postgres going undetected. I wouldn't tighten the readiness probe either, but I'd say this is about right.

PR #719, opened 2026-05-11
Add startup and liveness probes to dashboard

dspigarelli:
## What's changing and why?

Replaces the `wait-for-api-service` init container with startup/liveness probes on nginx (`httpGet /` on port `http`). Readiness still checks the API server health endpoint.


herrewig (approved):
LGTN

muditmathur11 (approved):

PR #723, opened 2026-05-12
Add blob service probes

dspigarelli:
# What's changing and why?

Adds startup, readiness, and liveness probes to the blob service container in the collector deployment, consistent with the pattern used by other services. Also exposes the HTTP port needed for the probes.

## How tested

Helm unit tests pass; snapshot updated.

### DIFF
--- a/charts/thoras/templates/collector/deployment.yaml
+++ b/charts/thoras/templates/collector/deployment.yaml
@@ -292,6 +292,30 @@ spec:
           limits: {{ .Values.metricsCollector.blobService.limits | toYaml | nindent 12 }}
           requests: {{ .Values.metricsCollector.blobService.requests | toYaml | nindent 12 }}
         {{- end }}
+        startupProbe:
+          httpGet:
+            path: /health
+            port: http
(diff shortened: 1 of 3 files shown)


charts/thoras/templates/collector/deployment.yaml
```diff
@@ -292,6 +292,30 @@ spec:
...
+            path: /health?db=true
+            port: http
+          periodSeconds: 10
+          failureThreshold: 3
+          timeoutSeconds: 2
+        livenessProbe:
+          httpGet:
+            path: /health
+            port: http
+          periodSeconds: 30
```
herrewig: Does this mean we have to wait half a minute before a second attempt if the first probe failed? if so, is that what we want?

dspigarelli: Do you have a recommendation for a better period? I could change to 10 second period (aligning with `readinessProbe`) and up the `failureThreshold` to 9.

herrewig (approved):
LGTM

PR #727, opened 2026-05-12
Add forecast worker scaling configuration

Rickyxstar:
# What's changing and why?

Adds optional feature flag `forecastWorkerScaler` for the worker service to dynamically scale the forecast worker deployment.

## What's changing

- New `featureFlags.forecastWorkerScaler` values block (disabled by default) with interval, deployment name/namespace, replica bounds, and ASTs-per-replica.
- Worker deployment env vars wired from the new values.
- Conditional `Role`/`RoleBinding` granting the worker `get/update/patch` on `deployments/scale` for the forecast worker deployment when the flag is enabled.

## How this helps customers

Lets the worker autoscale the forecast worker based on workload, improving forecast throughput without manual replica tuning.

## How tested

- `helm unittest ./charts/thoras --chart-tests-path ./charts/thoras/tests` (snapshot updated).

### DIFF
--- a/charts/thoras/values.yaml
+++ b/charts/thoras/values.yaml
@@ -57,6 +57,14 @@ featureFlags:
       forecastBlocks: "4h0s"
       # Forecast once per hour by default
       forecastCron: "17 * * * *"
+  forecastWorkerScaler:
+    enabled: false
+    interval: "1m"
+    deploymentName: "thoras-forecast-worker"
+    deploymentNamespace: "thoras"
+    minReplicas: 1
+    maxReplicas: 100
+    astsPerReplica: 67
   enableMigrationOnStartup: false
   enableCostSavingsSettingsRefresh: true
   enablePrometheusMetricScraping: false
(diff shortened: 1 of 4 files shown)


charts/thoras/values.yaml
```diff
@@ -57,6 +57,13 @@ featureFlags:
       forecastBlocks: "4h0s"
       # Forecast once per hour by default
       forecastCron: "17 * * * *"
+  forecastWorkerScaler:
+    enabled: false
+    interval: "1m"
+    deploymentName: "thoras-forecast-worker"
```
dspigarelli: I'd drop the `deploymentName` customization.

charts/thoras/values.yaml
```diff
@@ -57,6 +57,13 @@ featureFlags:
       forecastBlocks: "4h0s"
       # Forecast once per hour by default
       forecastCron: "17 * * * *"
+  forecastWorkerScaler:
```
dspigarelli: I would put this config in the `thorasForecast` section, not in feature flags.

dspigarelli (approved):
Mostly minor feedback.

PR #731, opened 2026-05-15
Remove unused forecast worker resource values

dspigarelli:
# What's changing and why?

Removes `resources`, `requests`, and `limits` fields from `thorasForecast` in `values.yaml` — these were unused dead config.

### DIFF
--- a/charts/thoras/values.yaml
+++ b/charts/thoras/values.yaml
@@ -397,16 +397,6 @@ thorasForecast:
   useAstMetricsSeries: false
   # Minimum lookback window required before autonomous scaling is enabled (minimum 3h).
   minLookbackToScale: "3h"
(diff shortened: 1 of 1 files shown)


charts/thoras/values.yaml
```diff
@@ -397,16 +397,6 @@ thorasForecast:
   useAstMetricsSeries: false
   # Minimum lookback window required before autonomous scaling is enabled (minimum 3h).
   minLookbackToScale: "3h"
-  # Specify the complete resources block (takes precedence if set)
```
dspigarelli, line 400: These were never used. Instead `thorasForecast.worker.requests`/etc. were the ones being used

garrettporter (approved):

PR #734, opened 2026-05-15
Default USE_AST_METRICS_SERIES to true

dspigarelli:
## What's changing and why?

Flips `useAstMetricsSeries` default to `true` so the forecast worker uses the new AST time series by default. Updates tests to match.


muditmathur11 (approved):
LGTM

PR #736, opened 2026-05-18
add oom remediation configurables to thoras-scaling-thoras workloads

garrettporter:
Relates to https://github.com/thoras-ai/platform/issues/4453

# What's changing and why?
Allows reactive OOM remediation for Thoras workloads when they are managed by Thoras.

### DIFF
--- a/charts/thoras/templates/worker/deployment.yaml
+++ b/charts/thoras/templates/worker/deployment.yaml
@@ -287,31 +287,31 @@ spec:
           - name: SERVICE_THORAS_SCALING_THORAS_METRICS_COLLECTOR_MEMORY_UPPER_BOUND
             value: {{ .Values.metricsCollector.upperbound.memory | quote }}
           - name: SERVICE_THORAS_SCALING_THORAS_FORECAST_WORKER_OOM_REMEDIATION_ENABLED
-            value: {{ .Values.thorasForecast.worker.oom_remediation.enabled | quote }}
+            value: {{ .Values.thorasForecast.worker.oom_remediation.enabled | ternary "true" "false" | quote }}
           - name: SERVICE_THORAS_SCALING_THORAS_FORECAST_WORKER_OOM_REMEDIATION_STABILIZATION_WINDOW
             value: {{ .Values.thorasForecast.worker.oom_remediation.stabilization_window | quote }}
           - name: SERVICE_THORAS_SCALING_THORAS_API_SERVER_OOM_REMEDIATION_ENABLED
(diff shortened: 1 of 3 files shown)


charts/thoras/templates/worker/deployment.yaml
```diff
@@ -286,6 +286,34 @@ spec:
             value: {{ .Values.metricsCollector.upperbound.cpu | quote }}
           - name: SERVICE_THORAS_SCALING_THORAS_METRICS_COLLECTOR_MEMORY_UPPER_BOUND
             value: {{ .Values.metricsCollector.upperbound.memory | quote }}
+          - name: SERVICE_THORAS_SCALING_THORAS_FORECAST_WORKER_OOM_REMEDIATION_ENABLED
+            value: {{ .Values.thorasForecast.worker.oom_remediation.enabled | ternary "true" "false" | quote }}
+          - name: SERVICE_THORAS_SCALING_THORAS_FORECAST_WORKER_OOM_REMEDIATION_STABILIZATION_WINDOW
+            value: {{ .Values.thorasForecast.worker.oom_remediation.stabilization_window | quote }}
```
dspigarelli: [sand] If value isn't overridden, then don't set the environment variable.
```suggestion
          {{- with .Values.thorasForecast.worker.oom_remediation.stabilization_window }}
          - name: SERVICE_THORAS_SCALING_THORAS_FORECAST_WORKER_OOM_REMEDIATION_STABILIZATION_WINDOW
            value: {{ . | quote }}
          {{- end }}
```

charts/thoras/values.yaml
```diff
@@ -126,6 +126,14 @@ thorasOperator:
   upperbound:
     cpu: ""
     memory: ""
+  # OOM remediation used when thorasManageThoras is enabled. Enabled by
+  # default so that an OOM on a Thoras workload raises its memory floor via
+  # the operator's standard OOM remediation flow. stabilization_window is a
+  # duration string (e.g. "1h"); empty defers to the operator default of
+  # max(forecast_interval, 1h).
+  oom_remediation:
```
dspigarelli, line 134: IMHO it's over-configuration allowing OOM remediation to be enabled and stabilization windows set per component.

dspigarelli (approved):
Changes are fine, minor [sand]-level comments.

charts/thoras/values.yaml
garrettporter: My thought was that since OOM remediation is configured per-AST that we should support the same for Thoras workloads.

PR #737, opened 2026-05-18
set thorasManageThoras default updateMode to in_place_or_recreate

garrettporter:
# What's changing and why?
By default, attempting an in-place resize if the cluster supports it, with a recreate callback.


herrewig (approved):
LGTM

PR #741, opened 2026-05-19
feat: add cert-manager support for webhook certs

Rickyxstar:

Lets customers who already run cert-manager provision the operator webhook TLS certs through it instead of the bundled `kube-webhook-certgen` Jobs. This aligns with their existing PKI/rotation workflows and removes the certgen Jobs, RBAC, NetworkPolicy, and registry pull secret when not needed.

## What's changing?

- Add `thorasOperator.webhookCertGen.certManager.enabled` (default `false`)
- When disabled: create a self-signed `Issuer` + `Certificate` producing the `thoras-tls-certs` secret and annotate the webhook configs with `cert-manager.io/inject-ca-from`
- When enabled: skip the certgen Job, ServiceAccount, Roles/Bindings, NetworkPolicy, and registry pull secret
- Operator deployment mounts `tls.crt`/`tls.key` keys (cert-manager's secret layout) in this mode

## How Tested

Added Helm unit tests covering both modes: cert-manager objects are absent by default, all certgen resources are skipped when enabled, webhook configs get the inject-ca annotation, and the operator volume references the correct secret keys.

### DIFF
--- a/charts/thoras/templates/operator/deployment.yaml
+++ b/charts/thoras/templates/operator/deployment.yaml
@@ -50,10 +50,17 @@ spec:
             defaultMode: 420
             secretName: thoras-tls-certs
             items:
+              {{- if .Values.thorasOperator.webhookCertGen.certManager.enabled }}
+              - key: tls.crt
+                path: tls.crt
+              - key: tls.key
+                path: tls.key
+              {{- else }}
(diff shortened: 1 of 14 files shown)


herrewig (approved):
LGTM

PR #747, opened 2026-05-22
Remove enableMigrationOnStartup feature flag

dspigarelli:
# What's changing and why?

The `enableMigrationOnStartup` flag has been permanently `true` and is no longer needed. Removing the flag, the now-dead `wait-for-postgres` init containers, and the related env vars from api-server, worker, and operator deployments.


muditmathur11 (approved):
✂️ ✂️ ✂️

PR #751, opened 2026-05-28
set thorasManageThoras default to autonomous

garrettporter:
# What's changing and why?
- Change the default thoras-scaling-thoras mode to autonomous so Thoras pods are automatically rightsized out-of-the-box.
- Comment that database forecasting should be migrated from cron based scheduling to interval based scheduling.

This PR can sit until we are ready to cut over to autoscaling Thoras workloads by default. I am not aware of any blockers for cutting over, but we should make the default behavior change clear to customers in release notes.


dspigarelli (approved):
Approving but not merging quite yet.

PR #753, opened 2026-06-02
feat: add USE_FORECASTER_COMPUTED_METRIC_ID env var to forecast worker

muditmathur11:
# What's changing and why?

Introducing new `USE_FORECASTER_COMPUTED_METRIC_ID` environment variable to the forecast worker, with possible values of true and false, defaulting to false.

When enabled alongside `USE_AST_METRICS_SERIES`, the forecaster computes legacy metric IDs locally in Python using the same hashing algorithm as Go's `MetricDefinition.Id()`, eliminating the catalog API call entirely. This is the final step in removing the forecaster's dependency on the catalog service.

[X] Do any feature flags need to be managed?

Yes, `thorasForecast.useForecasterComputedMetricId` in values.yaml (defaults to false, opt-in)

### DIFF
--- a/charts/thoras/README.md
+++ b/charts/thoras/README.md
@@ -148,6 +148,7 @@ All components support `<component>.useGlobalAffinity` (default: `true`) and `<c
 | thorasForecast.ignoreNewPods           | Boolean  | true                   | Directs forecaster to adjust CPU and memory metrics temporarily for new pods                   |
 | thorasForecast.enableDecoupledTraining | Boolean  | true                   | Enables async training mode where forecasts report "needs_training" instead of training inline |
 | thorasForecast.useAstMetricsSeries     | Boolean  | false                  | Enables catalog-free training data fetching via the AST metrics series endpoint                |
+| thorasForecast.useForecasterComputedMetricId | Boolean  | false             | Computes legacy metric IDs locally in Python instead of fetching from the catalog              |
 | thorasForecast.worker.podAnnotations   | Object   | {}                     | Pod Annotations for Thoras Forecast                                                            |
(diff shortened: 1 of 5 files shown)


charts/thoras/tests/forecast_worker_test.yaml
```diff
@@ -306,6 +306,25 @@ tests:
             name: USE_AST_METRICS_SERIES
             value: "false"
 
+  - it: renders USE_FORECASTER_COMPUTED_METRIC_ID as false by default
```
dspigarelli, line 309: [sand] The default value is already captured by the snapshot comparison test and I'm not sure the other is necessary.

charts/thoras/README.md
```diff
@@ -148,6 +148,7 @@ All components support `<component>.useGlobalAffinity` (default: `true`) and `<c
 | thorasForecast.ignoreNewPods           | Boolean  | true                   | Directs forecaster to adjust CPU and memory metrics temporarily for new pods                   |
 | thorasForecast.enableDecoupledTraining | Boolean  | true                   | Enables async training mode where forecasts report "needs_training" instead of training inline |
 | thorasForecast.useAstMetricsSeries     | Boolean  | false                  | Enables catalog-free training data fetching via the AST metrics series endpoint                |
+| thorasForecast.useForecasterComputedMetricId | Boolean  | false             | Computes legacy metric IDs locally in Python instead of fetching from the catalog              |
```
dspigarelli, line 151: [dust] My prettier formatter keeps the columns all aligned by adding whitespace.

dspigarelli (approved):

PR #754, opened 2026-06-02
feat: enable forecast worker scaling by default

Rickyxstar:

The forecast worker now scales its replicas automatically based on AST count, so customers' forecasting capacity keeps pace with workload growth without manual tuning. Previously this required opting in by editing values.

## What's changing?

- Set `thorasForecast.worker.scaler.enabled` to `true` by default in `values.yaml`
- Activates the worker's self-scaling RBAC (scale subresource Role/RoleBinding) and scaler env vars (interval, min/max replicas, astsPerReplica)

## How Tested

Config-only change; existing Helm unit tests cover the scaler templates. Verified the gated RBAC and deployment env vars render when enabled.


garrettporter (approved):

dspigarelli (approved):
Looks like the tests are failing, probably on snapshot deltas

PR #760, opened 2026-06-02
gate dashboard API calls via nginx allowlist

garrettporter:
Closes https://github.com/thoras-ai/platform/issues/4285

This change applies the principle of least privilege at the dashboard's reverse proxy: only the ~31 `{method, path}` combinations the dashboard actually uses are permitted; everything else returns `403`. It is not a full RBAC solution, but it raises the floor materially with no API server changes.

## What's changing?

- `charts/thoras/templates/dashboard/nginx-config-map.yaml` — add a `map $request_method$request_uri $thoras_dashboard_api_allowed` block enumerating allowed routes, and an `if ($thoras_dashboard_api_allowed != 1) { return 403; }` guard at the top of the existing `location /v1/` block. Default is deny.
- Allowlist sourced from a sweep of `dashboard/src` against the `@thoras/thoras-api` generated client plus two direct `axios`/`window.open` calls (`PATCH /v1/ast/.../mode`, `GET /v1/export`).

## How Tested

- `helm template` rendered the ConfigMap and confirmed the `map` block and the `if` guard appear inside `location /v1/`.
- Existing helm chart snapshot tests updated and pass.
- Regression tested by clicking around dashboard; no 403s seen
- From devtools console, verified 403s for API calls not in the allowlist

### DIFF
--- a/charts/thoras/templates/dashboard/nginx-config-map.yaml
+++ b/charts/thoras/templates/dashboard/nginx-config-map.yaml
@@ -22,6 +22,53 @@ data:
 
       include       mime.types;
 
+      # Least-privilege allowlist for dashboard -> API calls.
+      # Deny by default; only routes the dashboard actually uses are allowed.
+      # New dashboard API calls require a matching entry here or they will 403.
+      map $request_method$request_uri $thoras_dashboard_api_allowed {
+        default 0;
+
+        # GET — reads
+        ~*^GET/v1/ast(\?|$)                                                       1;
+        ~*^GET/v1/ast/[^/]+/[^/]+(\?|$)                                           1;
+        ~*^GET/v1/ast/[^/]+/[^/]+/pods(\?|$)                                      1;
+        ~*^GET/v1/ast/[^/]+/[^/]+/metrics/current(\?|$)                           1;
+        ~*^GET/v1/ast/[^/]+/[^/]+/suggestions(\?|$)                               1;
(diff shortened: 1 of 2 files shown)


charts/thoras/templates/dashboard/nginx-config-map.yaml
```diff
@@ -22,6 +22,53 @@ data:
...
+        ~*^GET/v1/views/namespaces/[^/]+/asts(\?|$)                               1;
+        ~*^GET/v1/views/namespaces/[^/]+/kpis(\?|$)                               1;
+        ~*^GET/v1/views/namespaces/[^/]+/labels(\?|$)                             1;
+        ~*^GET/v1/views/node-pools(\?|$)                                          1;
+        ~*^GET/v1/views/nodes(\?|$)                                               1;
+
+        # POST — dashboard-initiated mutations
+        ~*^POST/v1/ast/candidates/[^/]+(\?|$)                                     1;
+        ~*^POST/v1/ast/enroll/all(\?|$)                                           1;
+        ~*^POST/v1/ast/[^/]+/metrics/series(\?|$)                                 1;
```
dspigarelli: Technically this isn't a mutation. We're just using `POST` in order to pass in the list of metrics to fetch.

dspigarelli (approved):

charts/thoras/templates/dashboard/nginx-config-map.yaml
garrettporter: I moved it to its own section

PR #762, opened 2026-06-03
Feature gate setting AST owner references

dspigarelli:
# What's changing and why?

Adding a feature gate around setting the owner reference on AiScaleTargets to their target workload.


muditmathur11 (approved):

muditmathur11:
LGTM

PR #763, opened 2026-06-03
Enable AST owner references for all customers

dspigarelli:
# What's changing and why?

Enable automatically setting the owner reference on ASTs that reference their target workload.


muditmathur11 (approved):
LGTM

PR #766, opened 2026-06-03
ci: trigger release notes generation on helm chart release

muditmathur11:
# What's changing and why?

Adds two steps to the Release Charts workflow that fire after `chart-releaser` runs on a version bump.                                                   

The first step extracts the helm chart and platform versions from the diff automatically, no manual input needed. 
The second fires a `repository_dispatch` event to our private platform repository, which kicks off an internal workflow that generates a draft of the release notes for engineering review.                                                                                                                                    
                  
This requires a `PLATFORM_DISPATCH_TOKEN` secret in this repo, a fine-grained PAT scoped to `thoras-ai/platform` with Actions write permission, owned by the `thoras-ai org`. This has already been added to the repo secrets.
                                                                                                                                                         
The release process itself is unaffected. These steps run after `chart-releaser` completes, so a failure here has no impact on the release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,3 +40,45 @@ jobs:
         uses: helm/chart-releaser-action@v1.6.0
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+
+      - name: Extract versions for release notes
+        if: steps.version_bump.outputs.is_bump == 'true'
+        id: versions
+        shell: bash
+        run: |
+          LATEST_HELM=$(git diff HEAD^1..HEAD charts/thoras/Chart.yaml | grep '^+version:' | sed 's/^+version: *//')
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -40,3 +40,45 @@ jobs:
         uses: helm/chart-releaser-action@v1.6.0
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+
+      - name: Extract versions for release notes
+        if: steps.version_bump.outputs.is_bump == 'true'
+        id: versions
+        shell: bash
+        run: |
+          LATEST_HELM=$(git diff HEAD^1..HEAD charts/thoras/Chart.yaml | grep '^+version:' | sed 's/^+version: *//')
```
dspigarelli: [sand] This could use the value calculated in `version_bump`

dspigarelli (approved):

PR #767, opened 2026-06-03
feat: add PostgreSQL connection-pool settings with per-component overrides

Rickyxstar:

Operators can now tune PostgreSQL connection-pool behavior (max connections, idle time, lifetime) for the api-server, operator, and worker. Sensible defaults ship out of the box, and each component can override them independently to right-size pooling for larger clusters.

## What's changing?

- Add global `postgresql` defaults to `values.yaml` (`maxConns: 10`, `maxConnIdleTime: 5m`, `maxConnLifetime: 1h`)
- Add per-component `postgresql` override blocks for operator, api-server-v2, and worker
- Wire `SERVICE_POSTGRESQL_MAX_CONNS`, `_MAX_CONN_IDLE_TIME`, `_MAX_CONN_LIFETIME` env vars via `coalesce` (component override falls back to global default)

## How Tested

Added unit tests for each deployment (operator, api-server, worker) covering default values and per-component overrides. Updated global label snapshot. Run with `helm unittest ./charts/thoras --chart-tests-path ./charts/thoras/tests`.


dspigarelli (approved):
[dust] As discussed in person, I'd drop the specific default/coalesce tests and just stick with the snapshot tests.

PR #768, opened 2026-06-03
feat: add USE_FORECAST_ENRICHMENT env var to forecast worker

muditmathur11:
# What's changing and why?

Introducing new `USE_FORECAST_ENRICHMENT` environment variable to the forecast worker, with possible values of `true` and `false`, defaulting to `false`.

This gates per-metric enrichment computations (scalingClassification, burstProfile, evaluation) that run alongside core inference. These signals add latency to the forecast path on every metric for every AST on every cycle. Since they are not yet being acted on by clients, the overhead is not justified. This flag allows the implementation to remain in place while keeping inference fast until they are decoupled completely and ready to use by clients.


dspigarelli (approved):
[from in-person conversation] I'd drop the default/not-default tests and rely on the snapshot test to capture the default.

PR #769, opened 2026-06-03
Bump max connections to 20 from 10

dspigarelli:
# What's changing and why?

Bump default connections from 10 to 20 to give more default headroom to postgres pooling.


muditmathur11 (approved):
LGTM

PR #777, opened 2026-06-08
Remove obsolete thoras-monitor now that monitoring is in the worker

dspigarelli:
# What's changing and why?

With the last monitor being moved to the worker in https://github.com/thoras-ai/helm-charts/pull/756, the monitor is no longer necessary. In fact, it will cause problems if customers run with `thorasMonitor.enabled: true` because the monitor will start and immediately shutdown because it has no work to do.


muditmathur11 (approved):
LGTM

PR #778, opened 2026-06-08
Add feature flag for mirroring ASTs to the database

dspigarelli:
# What's changing and why?

Adding a feature flag to gate mirroring of ASTs to the database


muditmathur11 (approved):
looks good to me

PR #780, opened 2026-06-09
Default forecaster to generate legacy metric ids instead of using the catalog

dspigarelli:
# What's changing and why?

Switching a feature flag to cause the forecaster to generate legacy metric IDs instead of fetching them from the catalog.


muditmathur11 (approved):

muditmathur11:
looks good to me

Rickyxstar (approved):

PR #781, opened 2026-06-11
Conditionally drop anti-affinity between forecast workers and the collector

dspigarelli:
# What's changing and why?

With anti-affinities on the forecast worker and metrics collector, we prevent Karpenter consolidation.


herrewig (approved):
LGTM

PR #798, opened 2026-06-30
feat: allowlist CSV export endpoint in dashboard nginx

muditmathur11:
# What's changing and why?

Adds GET /v1/views/cost/details/export to the dashboard nginx allowlist. This endpoint was added to support CSV export of cost analysis data. Without this entry, the request is blocked with a 403 in deployed environments.

### DIFF
--- a/charts/thoras/templates/dashboard/nginx-config-map.yaml
+++ b/charts/thoras/templates/dashboard/nginx-config-map.yaml
@@ -44,6 +44,7 @@ data:
         ~*^GET/v1/views/asts/[^/]+/[^/]+(\?|$)                                    1;
         ~*^GET/v1/views/cost/analysis(\?|$)                                       1;
         ~*^GET/v1/views/cost/details(\?|$)                                        1;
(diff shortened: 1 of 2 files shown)


charts/thoras/templates/dashboard/nginx-config-map.yaml
```diff
@@ -44,6 +44,7 @@ data:
         ~*^GET/v1/views/asts/[^/]+/[^/]+(\?|$)                                    1;
         ~*^GET/v1/views/cost/analysis(\?|$)                                       1;
         ~*^GET/v1/views/cost/details(\?|$)                                        1;
+        ~*^GET/v1/views/cost/details/export(\?|$)                                1;
```
dspigarelli: Nit: whitespace

Image

dspigarelli (approved):
Just whitespace nit

charts/thoras/templates/dashboard/nginx-config-map.yaml
muditmathur11: Good eye!  🦅

PR #802, opened 2026-07-02
Upgrade timescale to 2.28.2

dspigarelli:
# What's changing and why?

Upgrading to the latest timescale image, 2.82.2-pg16


muditmathur11 (approved):
looks good

PR #808, opened 2026-07-08
Add model spec and suggestion status fields to AIScaleTarget CRD

Rickyxstar:
## What's changing and why?

Adds the `spec.model` block to the AIScaleTarget CRD for configuring forecasting and scaling behavior — `forecast_blocks`, `forecast_interval`/`forecast_cron`, `forecast_buffer_percentage`, `mode`, `percentile_value`, `min_lookback_to_scale`, and per-metric overrides — with CEL validation for mutual exclusivity and percentile constraints.

Also adds `status.horizontalSuggestion` and `status.verticalSuggestion` to expose forecast window, replica/resource recommendations, and data-sufficiency to consumers, and clarifies `disruption` docs to reflect that pod changes can be in-place resizes, not just evictions.

## How Tested

Ran `helm unittest ./charts/thoras` against the updated CRD schema.

### DIFF
--- a/charts/thoras/templates/crd/aiscaletarget.yaml
+++ b/charts/thoras/templates/crd/aiscaletarget.yaml
@@ -559,6 +559,111 @@ spec:
                     - observe
                     - observation
                     type: string
+                  model:
+                    description: |-
+                      Defines the forecasting and scaling behavior for an AIScaleTarget, including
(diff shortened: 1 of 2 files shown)


dspigarelli (approved):
Needs an update the AST template.

PR #809, opened 2026-07-10
Validation requiring cpu or memory for containers

dspigarelli:
# What's changing and why?

Vertical container specs must specify either a `cpu` or `memory` block. Otherwise, right-sizing doesn't know which dimensions are acceptable for resizing.


muditmathur11 (approved):
Looks good to me

PR #814, opened 2026-07-16
Remove initial DSA update policy

dspigarelli:
# What's changing and why?

We're not going to support `initial` update policy for DSA initially (no pun intended).


muditmathur11 (approved):
Looks good to me

PR #816, opened 2026-07-17
Remove 4.132.1 build

dspigarelli:
# What's changing and why?

The build was broken because the operator failed to come up unless `featureFlags.enableDaemonSetAutoscaler` is `true`


muditmathur11 (commented):
LGTM

PR #820, opened 2026-07-22
Allow dashboard service targetPort override

dspigarelli:
## What's changing and why?

Adds `thorasDashboard.service.targetPort` so the dashboard Service can route to an oauth2-proxy sidecar (e.g. port 4180) instead of the app container directly, enabling auth bypass via sidecar.

Defaults to `containerPort` when unset — no behavior change for existing installs.


muditmathur11 (approved):
Looks good to me

PR #821, opened 2026-07-22
Enable AST mirroring by default

dspigarelli:
# What's changing and why?

Testing of the mirroring of ASTs is complete in our test clusters.


muditmathur11 (approved):
LGTM

PR #822, opened 2026-07-22
Disable forecast worker anti-affinities

dspigarelli:
# What's changing and why?

Forecast workers can play on the same pods next to metrics-collectors.


muditmathur11 (approved):
LGTM

PR #827, opened 2026-07-28
Allow persistent volumes API in dashboard NGINX

Rickyxstar:
# What's changing and why?

Adds an NGINX allow-list rule for `GET /v1/persistent-volumes` so the dashboard can reach this endpoint. Without it, requests to the persistent volumes API are blocked by the reverse proxy, preventing related dashboard views from loading data.


muditmathur11 (approved):
LGTM

PR #828, opened 2026-07-28
Remove DSA feature flag

dspigarelli:
# What's changing and why?

Removing the feature flag for DaemonSetAutoscaling. The feature will continue to be gated strictly by the presence of DaemonSetAutoscaler resources


muditmathur11 (approved):
Looks good to me

PR #831, opened 2026-07-29
Increase startup recover window to 5 minutes

dspigarelli:
# What's changing and why?

While investigating a WAL corruption, I found the following feedback:

> The one legitimate concern with the 30-second timeout: a healthy database with a lot of crash-recovery WAL to replay (e.g., after an unclean shutdown with 10+ minutes of WAL) could legitimately take >30 seconds to start. It would trigger a false pg_resetwal -f. Worth considering whether 60 or 120 seconds is safer for production deployments.

Originally this had been 5 minutes, increasing the default window to 300s.

### DIFF
--- a/charts/thoras/templates/collector/deployment.yaml
+++ b/charts/thoras/templates/collector/deployment.yaml
@@ -105,20 +105,37 @@ spec:
         args:
           - |
             echo "Checking WAL integrity..."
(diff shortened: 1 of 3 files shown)


muditmathur11 (approved):
Looks good

PR #834, opened 2026-07-30
Pass namespace to operator and worker

dspigarelli:
# What's changing and why?

The worker and operator both need the thoras namespace in order to read Thoras service config maps


muditmathur11 (approved):
LGTM

PR #836, opened 2026-07-30
Use ClusterRoles for querying Nodes, PVs and StorageClasses, even when scoped to specific namespaces

dspigarelli:
# What's changing and why?

When scoping Thoras to specific namespaces, ala `rbac.namespace`, we still need ClusterRoles, not simply Roles to query Nodes, PVs, and StorageClasses.


Rickyxstar (approved):

charts/thoras/tests/api_service_deployment_rbac_test.yaml
```diff
@@ -1,27 +0,0 @@
...
-  thorasApiServerV2:
-    rbac:
-      create: true
-tests:
-  - it: Default matches snapshot
-    chart:
-      version: "4.7.0"
-    asserts:
-      - matchSnapshot: {}
-  - it: thoras role includes configmap permissions
```
dspigarelli, line 17: This test isn't even remotely useful.

PR #839, opened 2026-08-05
chore: increase default min lookback to scale and wire worker templat

muditmathur11:
# What's changing and why?

- Bumps `thorasForecast.minLookbackToScale` default from 3h to 24h — gives workloads a full day of history before autonomous scaling kicks in, reducing churn on fresh installs
- Wires `SERVICE_MIN_LOOKBACK_TO_SCALE` into the worker deployment template — was previously missing, so the helm value was silently ignored by the worker; per-workload overrides via AST or CASTT were unaffected


dspigarelli (approved):
I'm not opposed to this change, I'm just  not aware of the decisions behind the change.

PR #845, opened 2026-08-10
feat: add configurable idle utilization threshold

Rickyxstar:
# What's changing and why?

Lets customers tune the utilization level at which a workload is classified as idle, rather than relying on a hardcoded server-side value.

## What's changing?

- Add `utilizationThresholds.idle` (default `5`) to `values.yaml`
- Pass it as `SERVICE_IDLE_THRESHOLD` to the API server v2 deployment
- Update global label snapshot

## How Tested

`helm unittest ./charts/thoras` — 265 tests / 68 snapshots pass, including the updated snapshot asserting the new env var.

## Note

Chart version in `Chart.yaml` is not bumped, so merging this won't trigger a release.


muditmathur11 (approved):
Looks good to me

herrewig (commented):
LGTM

PR #848, opened 2026-08-12
feat: allow node summary endpoint in dashboard nginx config

Rickyxstar:

Unblocks the node summary view in the dashboard. Without this allowlist entry, nginx rejects `GET /v1/views/nodes/summary`, so the UI can't load node summary data.

## What's changing?

- Add `~*^GET/v1/views/nodes/summary(\?|$)` to the dashboard nginx allowlist
- Refresh the `dashboard_configmap` snapshot to match

## How Tested

`helm unittest ./charts/thoras` — 265 tests / 68 snapshots pass. The snapshot diff is limited to the single added allowlist line.


herrewig (approved):
LGTM

PR #850, opened 2026-08-13
feat: add extraIngressRules to all component network policies

Rickyxstar:

With `networkPolicy.enabled`, the default ingress rule only admits pods in the Thoras namespace, so cross-namespace Prometheus scrapes are dropped — the chart advertises `prometheus.io/scrape` annotations and `http-metrics` ports that nothing outside the namespace can reach. Customers can now allow their monitoring stack in without disabling network policies wholesale.

## What's changing?

- `extraIngressRules` on all six components (operator, collector, api-server-v2, worker, dashboard, forecast-worker), for both `kubernetes` and `cilium` flavors
- Mirrors the existing `extraEgressRules` pattern; defaults to `[]`
- Rules append after built-in ingress, so dashboard 8080 and operator webhook 9443 are unaffected

Example — letting Prometheus scrape the operator on 9101:

```yaml
thorasOperator:
  extraIngressRules:
    - from:
        - namespaceSelector:
            matchLabels:
              kubernetes.io/metadata.name: monitoring
      ports:
        - port: 9101
          protocol: TCP
```

Note metrics ports differ per component (operator/forecast-worker 9101, worker 9102). Prometheus also needs matching egress if it has its own policy.

## How Tested

Added 2 unit tests (kubernetes + cilium injection); full suite passes at 267 tests / 68 snapshots. Rendered all 6 components x 2 flavors with a probe rule and asserted via YAML parsing that it lands in `spec.ingress`, not `spec.egress`. Diffed rendered output against main with empty defaults — network policies byte-identical, so no-op for existing installs.

### DIFF
--- a/charts/thoras/templates/api-server-v2/network-policy.yaml
+++ b/charts/thoras/templates/api-server-v2/network-policy.yaml
@@ -18,6 +18,9 @@ spec:
   # Allow all ingress from pods within the same namespace
   - from:
     - podSelector: {}
+  {{- with .Values.thorasApiServerV2.extraIngressRules }}
+  {{- toYaml . | nindent 2 }}
+  {{- end }}
   egress:
   # Allow all egress to pods within the same namespace
   - to:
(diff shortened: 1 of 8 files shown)


herrewig (approved):
LGTM

PR #851, opened 2026-08-13
Feature flag for creating secret-based authn

herrewig:
# What's changing and why?

Give users the ability to opt-in to authn enforcement on the api


muditmathur11 (approved):
LGTM

PR #852, opened 2026-08-13
fix: make API server netpol ports configurable

Rickyxstar:
# What's changing and why?

Clusters that front the Kubernetes API on a non-standard port (minikube uses 8443) couldn't install with network policies enabled — the hardcoded 443/6443 egress rule blocked the operator, worker, api-server, and certgen from reaching the API. Ports are now configurable via `networkPolicy.apiServerPorts`, defaulting to the previous 443/6443 so existing installs are unaffected.

- Add `networkPolicy.apiServerPorts` to values.yaml
- Extract the duplicated egress rule into a `thoras.apiServerEgressRule` helper used by all four `kubernetes`-flavor policies
- Empty list drops the rule entirely rather than rendering a portless rule (which would allow egress on *every* port)
- Cilium flavor is unaffected — it targets `toEntities: [kube-apiserver]` and needs no ports

## How Tested

Added 3 helm unittest cases: custom port list applies to all four policies, empty list omits the rule, and cilium flavor ignores the setting. Full suite passes (270 tests).

### DIFF
--- a/charts/thoras/templates/_helpers.tpl
+++ b/charts/thoras/templates/_helpers.tpl
@@ -184,6 +184,29 @@ true
 {{- end -}}
 
 {{/*
+Egress rule allowing components to reach the Kubernetes API server, for the
+"kubernetes" NetworkPolicy flavor.
+
+Standard NetworkPolicy cannot target the API server by label, so this permits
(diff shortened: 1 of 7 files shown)


muditmathur11 (approved):
Looks good to me

PR #854, opened 2026-08-13
Forward auth secret through dashboard proxy

dspigarelli:
## What's changing
Wires the dashboard's nginx proxy to forward the existing shared auth secret as an `Authorization: Bearer` header on its `/v1/*` calls to api-server-v2, so enabling `featureFlags.enableSimpleAuthSecret` doesn't break the dashboard. Injected server-side via nginx's envsubst mechanism — secret never touches the ConfigMap, Helm history, or the browser.

## How this helps the customer
Lets operators turn on shared-secret auth without losing dashboard functionality.

## Testing
`helm unittest` (276/276 passing) + verified against the real `nginx:alpine` image (non-root UID, entrypoint/envsubst behavior, secret substitution).


herrewig (approved):
LGTM

PR #856, opened 2026-08-13
Enable simple auth by default

dspigarelli:
# What's changing and why?

Enabling simple auth between Thoras services by default


muditmathur11 (approved):
👍

herrewig (approved):
LGTM

PR #858, opened 2026-08-13
Enable turning off dashboard

dspigarelli:
# What's changing and why?

The `thorasDashboard.enabled` was only partially applied.


muditmathur11 (approved):
LGTM

PR #861, opened 2026-08-13
feat: allow pod logs route through dashboard API proxy

Rickyxstar:
# What's changing and why?

The dashboard's nginx proxy denies by default, so its pod-logs request was returning 403. This allowlists the route, letting users read pod logs from the UI instead of dropping to `kubectl`.

## What's changing?

- Allow `GET /v1/pods/{namespace}/{pod}/logs` in the dashboard nginx allowlist
- Update `dashboard_configmap_test.yaml` snapshots

## How Tested

`helm unittest ./charts/thoras` — 281 tests / 68 snapshots pass.

Checked the pattern against the caller's URL shape (`/v1/pods/${encodeURIComponent(ns)}/${encodeURIComponent(pod)}/logs`): bare and query-string forms (`?container=`, `?tailLines=`, `?previous=`) allow; `[^/]+` keeps percent-encoded segments (`%2F`) inside one path segment, and `/logs/extra`, `/logsfoo`, and `..` traversal all still deny.


muditmathur11 (approved):
LGTM

